PR #9, opened 2020-05-13
Add welcome/depart messages

naomi-lgbt:
Did some research on how to write this - I assume it goes in the same file where message commands are read?

Might need to be tested -  I don't have the bot set up locally yet.

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -7,11 +7,11 @@ module.exports = function bootstrap(client) {
   const commands = [eightball];
   
   client.on('guildMemberAdd', member => {
-    member.guild.channels.get('channelID').send("**" + member.user.username + "** has joined the server! :)")
+    member.guild.channels.cache.find(ch => ch.name === 'introduction').send("**" + member.user.username + "** has joined the server! :)")
   })
             
   client.on('guildMemberRemove', member => {
-    member.guild.channels.get('channelID').send("**" + member.user.username + "** has left the server! :(")
+    member.guild.channels.cache.find(ch => ch.name === 'introduction').send("**" + member.user.username + "** has left the server! :(")
   })
   
   client.on('message', (message) => {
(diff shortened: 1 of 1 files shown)


bradtaniguchi:
> Might need to be tested - I don't have the bot set up locally yet.

Yea it would be preferred to get it working first locally to test it. Need to add some info on at least where to get this information.

naomi-lgbt:
I'll fiddle around and see if I can get the bot set up locally and report back.

naomi-lgbt:
image

Local testing discovered incorrect syntax with initial commit. Discord.js documentation demonstrated a method for sending a welcome message to a specific channel - in my case, "introduction". We can change the target channel if we want.

naomi-lgbt:
https://discord.js.org/#/docs/main/stable/examples/greeting

bradtaniguchi:
Thinking about the original gitter bot, it never had a welcome system like this as Quincy said it would spam to the channel to much.

This bot should be able to **message a person directly after they have joined**, and have a formatted message (so we can link stuff, like the freeCodeCamp code of conduct).

As such, I **dont** think we want this feature right now. Its good for experimenting, but this bot will at least only have features that could potentially go into the future "official" bot.

naomi-lgbt:
image
Direct welcome message tested, appears functional.

Ryuno-Ki:
Could you try adding some Markdown? Like a link or italic?

naomi-lgbt:
New welcome message, with a link and an emoji for @Ryuno-Ki 😄 
Also added the `help` requirement to match this file to #15

bradtaniguchi (approved):
Looks good to me!

**edit** just noticed this has an announcement message logic. I don't think we need/want this in the FCC version of this bot, but I'll accept it for now because
1. I already said LGTM haha
2. We might want to perform "teardown" logic of any account connection between the user's discord account and FCC in the future.
3. This bot can spam a bit, no biggie.

Issue #7, opened 2020-05-13
Finalize "eightBall" command 


bradtaniguchi:
The only available command is the "eightBall" command available at `!prefix`. It doesn't currently work and can actually be implemented.

PR #12, opened 2020-05-13
Set up eightball command

naomi-lgbt:
To address #7:
Created array with the 20 responses from the original Magic 8 Ball.

### DIFF
--- a/src/commands/eightball.js
+++ b/src/commands/eightball.js
@@ -8,7 +8,11 @@ module.exports = {
    */
   command: function eightBall(message) {
     console.log('eightBall', message);
+    let options = ["As I see it, yes." , "Ask again later." , "Better not tell you now.", "Cannot predict now.", "Concentrate and ask again.", "Don't count on it.", "It is certain.", "It is decidedly so.", "Most likely.", "My reply is no.", "My sources say no.", "Outlook not so good.", "Outlook good.", "Reply hazy, try again.",
+"Signs point to yes.", "Very doubtful.", "Without a doubt.", "Yes.", "Yes - definitely.", "You may rely on it."]
+    let rand = Math.floor(Math.random() * 20)
     // TODO:
-    message.channel.send('magic not ready yet');
+    message.channel.send(options[rand]);
   },
 };
+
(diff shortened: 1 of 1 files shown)


naomi-lgbt:
image
Tested locally.

naomi-lgbt:
image
Works on another channel as well.

Twaha-Rahman:
LGTM!

src/commands/eightball.js
```diff
@@ -8,7 +8,11 @@ module.exports = {
    */
   command: function eightBall(message) {
     console.log('eightBall', message);
+    let options = ["As I see it, yes." , "Ask again later." , "Better not tell you now.", "Cannot predict now.", "Concentrate and ask again.", "Don't count on it.", "It is certain.", "It is decidedly so.", "Most likely.", "My reply is no.", "My sources say no.", "Outlook not so good.", "Outlook good.", "Reply hazy, try again.",
+"Signs point to yes.", "Very doubtful.", "Without a doubt.", "Yes.", "Yes - definitely.", "You may rely on it."]
+    let rand = Math.floor(Math.random() * 20)
```
bradtaniguchi: Change `let` to `const` for both variable declarations. Generally don't use `let` if the variables aren't going to change.

src/commands/eightball.js
```diff
@@ -8,7 +8,11 @@ module.exports = {
...
     // TODO:
```
bradtaniguchi: Remove TODO

src/commands/eightball.js
bradtaniguchi: Change the `20` to `options.length`, this way if we add more options to the list we wont have to manually update the count.

bradtaniguchi (changes requested):
Update a few things to help make the code easier to maintain.

src/commands/eightball.js
naomi-lgbt: Oops. :)

src/commands/eightball.js
naomi-lgbt: That's a good idea - I didn't anticipate that.

naomi-lgbt: Makes sense to me.

naomi-lgbt:
Proposed changes have been implemented and new commit generated.

bradtaniguchi (approved):
LGTM :+1:

PR #15, opened 2020-05-13
Patch 3

naomi-lgbt:
Adding help command and updating bootstrap file to run it.

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -4,7 +4,7 @@ const help = require('./help')
  * Bootstraps all commands to the client.
  */
 module.exports = function bootstrap(client) {
-  const commands = [eightball];
+  const commands = [eightball, help];
 
   client.on('message', (message) => {
     for (let command of commands) {
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -1,5 +1,5 @@
 const eightball = require('./eightball');
-
+const help = require('./help')
 /**
  * Bootstraps all commands to the client.
  */
--- a/src/commands/help.js
+++ b/src/commands/help.js
@@ -38,7 +38,7 @@ module.exports = {
 
       message.author.send(helpEmbed);
     } catch (error) {
-      console.log(error);
+      console.error(error);
     }
   },
 };
--- a/src/commands/help.js
+++ b/src/commands/help.js
@@ -1,6 +1,6 @@
-const Discord = require('discord.js')
+const Discord = require("discord.js");
 module.exports = {
-  prefix: '!help',
+  prefix: "!help",
   /**
    * @name help
    * Displays currently available commands.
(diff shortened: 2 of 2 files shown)


src/commands/bootstrap.js
```diff
@@ -1,5 +1,5 @@
 const eightball = require('./eightball');
-
+const help = require('./help')
```
bradtaniguchi: This is imported, but not added to the list of commands, so this command wont be available.

src/commands/help.js
```diff
@@ -0,0 +1,35 @@
...
+   */
+
+  /**
+   *Use this format to add new commands:
+   *{name: "Command Name", value: "Command Description"},
+   */
+
+  command: function help(message) {
+    console.log('help', message);
+      const helpEmbed = new Discord.MessageEmbed()
```
bradtaniguchi: Wrap a `try{} catch(){}` around the sending of the message so the bot doesn't crash.
in the `catch` be sure to log the error.

src/commands/help.js
```diff
@@ -0,0 +1,35 @@
...
+      .setColor('#0099ff')
+      .setTitle('Bot Information')
+      .setDescription("Hello! I am a test bot created by brantaniguchi and members of FreeCodeCamp.org to experiment with the process of building a Discord Bot. You can view my source code at https://github.com/bradtaniguchi/discord-bot-test")
+      .addFields(
+        { name: '\u200B', value: '\u200B' },
+        {name: "Commands", value: "Here are my currently available commands!"},
+        {name: "!eightball <optional string>", value: "Returns a response from a magic 8 ball toy"},
```
bradtaniguchi: The magic 8 ball is not a toy! Its a magic pool ball!

src/commands/help.js
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * @param {Discord.Message} message the message provided
+   */
+
+  /**
+   *Use this format to add new commands:
+   *{name: "Command Name", value: "Command Description"},
+   */
+
+  command: function help(message) {
```
bradtaniguchi: Formatting for this code doesn't look right, as the indentation is off. Are you running this code in an editor that can format with prettier on save? Once automated linting is setup this will error out.

bradtaniguchi (changes requested):

src/commands/bootstrap.js
naomi-lgbt: Oops. Missed that line.

src/commands/help.js
nhcarrigan: Ran it through prettier!

src/commands/help.js
naomi-lgbt: Added a catch feature

src/commands/help.js
naomi-lgbt: FIXED

src/commands/help.js
naomi-lgbt: Ran through prettier

src/commands/bootstrap.js
naomi-lgbt: Fixed

src/commands/help.js
```diff
@@ -0,0 +1,46 @@
...
+        //for spacing, leave this one at bottom.
+        { name: "\u200B", value: "\u200B" }
+      )
+      .setFooter("I am not affiliated with FreeCodeCamp in any way.");
+
+    message.author
+      .send(helpEmbed)
+      .catch((err) => message.channel.send(err.message));
+    message.channel.send(message.author.tag + ", I've sent you a message!");
+  },
```
bradtaniguchi: Instead of using `.catch` we should use `async/await` in a `try {} catch() {}`. I'm on the fence about sending a message with the error, id rather log it and just send nothing back to the user.

src/commands/help.js
```diff
@@ -0,0 +1,46 @@
...
+   * @name help
+   * Displays currently available commands.
+   *
+   * @param {Discord.Message} message the message provided
+   */
+
+  /**
+   *Use this format to add new commands:
+   *{name: "Command Name", value: "Command Description"},
+   */
```
bradtaniguchi: Combine these two comments into 1

naomi-lgbt: Done :)

src/commands/help.js
naomi-lgbt: Okay, I believe I implemented this change correctly?

bradtaniguchi (commented):
Make 1 change, then LGTM!

src/commands/help.js
```diff
@@ -0,0 +1,44 @@
...
+          //new command lines go below here.
+
+          //for spacing, leave this one at bottom.
+          { name: "\u200B", value: "\u200B" }
+        )
+        .setFooter("I am not affiliated with FreeCodeCamp in any way.");
+
+      message.author.send(helpEmbed);
+    } catch (error) {
+      console.log(error);
```
bradtaniguchi: Change to `console.error` instead of `console.log`

naomi-lgbt: Done!
Heh - I've never actually heard of console.log alternatives until today 😬

Ryuno-Ki: https://developer.mozilla.org/en-US/docs/Web/API/Console#Methods

Issue #17, opened 2020-05-13
Add feature to format and prettify a code-block


bradtaniguchi:
Given a message link/ref the bot should be able to do the following:
1. Determine what language it is (among supported list, determined)
  - probably a library out there **somewhere** that can help with this. Obviously if a message has text **and** code, the bot probably wont be able to differentiate. 
2. Format the code in backticks with the correct syntax highlighting using something like prettier to format the code sensibly
3. If at any point the bot can't format/understand the language, return a "sorry" like message.

This feature should only be called with the format similar to:
```
!format <message_link>
```

This shouldn't be ran all the time as this can be a CPU intensive task if ran all the time.

We also need to make sure we **don't** perform remote code execution, so we need libs well respected libs to handle the code formatting.

This command might also work just to format a given text block without a link to another message. 

**edit** we might be able to use the reactions feature to label code to "format":
https://discordjs.guide/popular-topics/reactions.html#awaiting-reactions

Ryuno-Ki:
> probably a library out there somewhere that can help with this. Obviously if a message has text and code, the bot probably wont be able to differentiate.

GitHub itself is using [Linguist](https://github.com/github/linguist) for that.

Ryuno-Ki:
> we might be able to use the reactions feature to label code to "format"

One idea was to use :robot: emoji to trigger them.

bradtaniguchi:
> GitHub itself is using Linguist for that.

Linguist looks pretty heavy duty, and would require a ruby runtime which would increase the pre-req for this project beyond just node. Optimally we'd find a respectable node lib that we can leverage as-is, or piggieback on whatever the formatters use (if they work that way hehe)

Twaha-Rahman:
Let's narrow down our target for now. We can start by creating an MVP release, which will format all web related languages. (For example -> `JavaScript,  TypeScript,  Flow, JSX, JSON,
CSS, SCSS, Less,
HTML, Vue, Angular,
GraphQL, Markdown, YAML`) All of these can be formatted using Prettier!

I am suggesting this because the majority of people in the FCC Discord are web devs and they share within these formats.

After we get the MVP release out, then we can focus on adding other languages too like-
1. Python
2. PHP
etc....

PR #27, opened 2020-05-16
Add format command to the bot

Twaha-Rahman:
Since this a **really** big PR, I thought it was a good idea to do PR and keep pushing to it so others can keep reviewing the code, while I finish up the remaining thing.

State of the command :

- [x] Supports add within code block (for any language) [Adding codes inside three backticks]
- [x] Supports syntax highlighting (sometimes have trouble detecting language, but overall looks promising)
- [x] Supports formatting code block with Prettier (supports only HTML, CSS, JS, JSON, Markdown)

Please review the PR and suggest me any improvement!


Twaha-Rahman:
This PR will address #17 too.

Twaha-Rahman:
The tests in Travis aren't passsing due to #28.

src/commands/add-formatting.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if (
+      detectedWithLanguageClassifier === detectedWithHLJS.language ||
+      detectedWithHLJS.top.aliases.includes(detectedWithLanguageClassifier) // need compare it with hljs's second guess here too
+    ) {
+      let supportedLanguage = isSuppotedByPrettier(
+        detectedWithLanguageClassifier
+      );
+
+      if (supportedLanguage) {
```
bradtaniguchi: This `if else` call splits the function into two distinct paths. We could use the early return:
```
if (supportedLanguage) {
  //..
  return;
}
// the code from the else side
```

src/commands/add-formatting.js
```diff
@@ -0,0 +1,134 @@
...
+        const formattedCode = formatter(content, supportedLanguage);
+        message.channel.send(
+          `\`\`\`${detectedWithLanguageClassifier}\n${formattedCode}\`\`\``
```
bradtaniguchi: This is nasty, it should be abstracted away into a function so the anywhere we need this formatted code, we can just call the utility function.

src/commands/add-formatting.js
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+      const reactionEmojies = ['🙂', '🙃', '😃'];
+
+      const confusedMessage = confusedMessageGenerator(
+        languageGuesses,
+        reactionEmojies
+      );
+
+      const refToPromptMsg = message.channel.send(confusedMessage);
+      refToPromptMsg.then((ref) => {
```
bradtaniguchi: This callback can probably be abstracted away into another file/function. Not sure what to name the function/file yet tho.

src/commands/add-formatting.js
```diff
@@ -0,0 +1,134 @@
...
+        message.channel.send(
+          `\`\`\`${detectedWithLanguageClassifier}\n${formattedCode}\`\`\``
+        );
+      } else {
+        message.channel.send(
+          `\`\`\`${detectedWithLanguageClassifier}\n${content}\`\`\``
+        );
+      }
+    } else {
+      const reactionEmojies = ['🙂', '🙃', '😃'];
```
bradtaniguchi: Instead of using an array here and then referencing each emoji via their index, we can use an enum-like approach (the keys of the "enum" can be named to whatever makes sense, or match with their actual name):
```javascript
const reactioEmojies = {
  good: '🙂',
  bad: '🙃',
  great: '😃'
};
//..
switch(reactionEmoji) {
  case 'good':
 //..
```

src/commands/add-formatting.js
```diff
@@ -0,0 +1,134 @@
+const lang = require('language-classifier');
+const hljs = require('highlight.js');
```
bradtaniguchi: So whats the differences between these two again and the benefit we get from using both rather then just 1 or the other?

src/utilities/confusedMessageGenerator.js
```diff
@@ -0,0 +1,17 @@
+function confusedMessageGenerator(possibleLanguages, reactionEmojies) {
```
bradtaniguchi: rename this file to kabab case to match with the other files and export it on the same line, again to match the other files so far.

Add comments/docs on what each of the params are, represent and their data types.

src/utilities/confusedMessageGenerator.js
```diff
@@ -0,0 +1,17 @@
...
+  let questions = [];
+
+  possibleLanguages.forEach((language, index) => {
+    questions.push(
+      `Is it \`${language}\`? (If it is \`${language}\`, then react to the unformatted message with a ${reactionEmojies[index]} emoji)`
+    );
+  });
```
bradtaniguchi: This can be done using a `.map` call rather than a `.forEach`

src/utilities/formatter.js
```diff
@@ -0,0 +1,46 @@
+const prettier = require('prettier');
+const stripAnsi = require('strip-ansi');
+
+function formatter(unformattedCode, language) {
```
bradtaniguchi: export this on the first line, and add comments

src/utilities/formatter.js
```diff
@@ -0,0 +1,46 @@
...
+  console.log(options);
+
+  let formattedCode = '';
+
+  try {
+    formattedCode = prettier.format(unformattedCode, options);
+  } catch (error) {
+    formattedCode = stripAnsi(
+      `SyntaxError: Unexpected token (${error.loc.start.line}:${error.loc.start.column})\n\n${error.codeFrame}`
+    );
```
bradtaniguchi, line 58: What does the stripAnsi do and why is it needed? This looks like magic to me atm.

src/utilities/isSuppotedByPrettier.js
```diff
@@ -0,0 +1,45 @@
+function isSuppotedByPrettier(languageName) {
```
bradtaniguchi: export on first line, rename file and add comments

bradtaniguchi (changes requested):
Reviewed the code, most of the comments are either me just being confused, and pointing out changes that can be made for code quality (and to help me from being confused hehe)

src/commands/add-formatting.js
Twaha-Rahman: `hljs` is the one which supports many language and that is why it sometimes confuses simple JS/CSS/HTML codes for other languages that nobody ever heard of. (Overall `hljs` is good but it can sometimes suffer from a little accuracy problem) As for `language-classifier`, it only supports a handful of languages and quite accurate when it is inside it's comfort zone but panic when it finds languages it doesn't support.

Another thing to note here is, `hljs` gives two guesses and `language-classifier` gives only one guess. Normally, `hljs` is the one which most of the times detects correctly but it sometimes can't properly guess simple things like CSS and HTML, that's where `language-classifier` comes in.

Overall, I think a **lot** testing needs to be done.

bradtaniguchi: I looked a little into the `language-classifier` package, and it looks like it hasn't been maintained for **years**, nor can I find any source code, only the 0.0.1 version on npm (which interestingly, isn't mentioned in the package.json)

due to this, it might make sense to focus on using **just** the hljs lib (which also hasn't been added to the package.json) as its at least maintained, with a **ton** of community support in comparison.

src/commands/add-formatting.js
Twaha-Rahman: Well spotted, should've used the modified version of guard clause...

src/commands/add-formatting.js
Twaha-Rahman: I always knew, I will forget about this.... I thought I will do it just before the PR.... Sorry that you had to read this horrible code...😥

src/commands/add-formatting.js
Twaha-Rahman: Done..... fixed it!

src/utilities/confusedMessageGenerator.js
Twaha-Rahman: Done renaming the file and adding JSdoc. Do you also want me to rename the function name to `kabab case`?

src/utilities/confusedMessageGenerator.js
Twaha-Rahman: Replaced `forEach` with `map`.

src/utilities/formatter.js
Twaha-Rahman: Done renaming the file and adding JSdoc.

src/utilities/isSuppotedByPrettier.js
Twaha-Rahman: Done!

src/commands/add-formatting.js
Twaha-Rahman: Done doing so!

src/utilities/formatter.js
Twaha-Rahman: When formatting code if Prettier's parser detects a syntax error, it stops parsing and shows an error by calling `console.error`. But I catch that error and send it back to Discord as message from the bot. But the problem is that Prettier _assumes_ that their error message will be displayed only in the console, so to help the devs figure out where a syntax error occurred, they colorize their error message with ANSI. This isn't a problem when it comes to console (who doesn't like colorful errors?) but if I send it to Discord, it can't display it. That's why we have to use `stripAnsi` to remove ANSI from the error message.

Shout out to @Ryuno-Ki as he helped me solve this problem as he previously ran into it where as I had no clue what went wrong!

src/commands/add-formatting.js
Twaha-Rahman: Yeah....I just checked it too. I think we should scrap `language-classifier` and just use `hljs`.

As for the packages not being linked in the `package.json`, I have fixed that.

Twaha-Rahman: Removed `language-classifier` compleletely..

src/commands/add-formatting.js
```diff
@@ -0,0 +1,62 @@
...
+const confusedMessageGenerator = require('../utilities/confused-message-generator');
+const processReply = require('../utilities/process-reply');
+
+module.exports = {
+  prefix: '',
```
bradtaniguchi: How is this command initially called? Or is it done via some emoji reaction on any message?

If this command is called on emoji reaction we could just not use this prefix, and handle this command differently then the rest of the commands.

src/commands/add-formatting.js
```diff
@@ -0,0 +1,62 @@
...
+      secondOption: '🙃'
+    };
+
+    const reactionEmojies = [];
+
+    const objectKeys = Object.keys(reactionOptionsObj);
+
+    objectKeys.map((key) => {
+      reactionEmojies.push(reactionOptionsObj[key]);
+    });
```
bradtaniguchi: This looks like we are getting the **list of values** from the `reactionObjectsObj`, if such is the case, these lines can be condensed to just:
```js
const reactionEmojis = Object.values(reactionOptionsObject):
```

PS. there is also the method `Object.entries` which will return an array of `[key, value]` from the object, which could of been used if we needed the `key` and `value`, without doing a "lookup". We don't need the `key` in this case, so we can just use `Object.values` this time :)

src/commands/add-formatting.js
```diff
@@ -0,0 +1,62 @@
...
+      reactionEmojies,
+      reactionOptionsObj,
+      languageGuesses,
+      content
+    };
+
+    const refToPromptMsg = message.channel.send(confusedMessage);
+    refToPromptMsg.then((ref) => {
+      processReply({ ...requiredInfoObj, ref });
+    });
```
bradtaniguchi: change to async/await to match with most of the other code in the project.

src/utilities/confused-message-generator.js
```diff
@@ -0,0 +1,29 @@
...
+  possibleLanguages,
+  reactionEmojies
+) {
+  let questions = [];
+
+  possibleLanguages.map((language, index) => {
+    questions.push(
+      `Is it \`${language}\`? (If it is \`${language}\`, then react to the unformatted message with a ${reactionEmojies[index]} emoji)`
+    );
+  });
```
bradtaniguchi: Don't use a `.map` if you don't use the return value, also if your using it to push values into another array, you could just use `.map` like so:
```js
const questions = possibleLanguages.map((language, index) => `Is it \`${language}\`? (If it is \`${language}\`, then react to the unformatted message with a ${reactionEmojies[index]} emoji)`);
```

src/utilities/confused-message-generator.js
```diff
@@ -0,0 +1,29 @@
...
+
+  let confusedMessage = `${questions.join(
+    '\n'
+  )}\n\n(P.S.: React to this message...)`;
```
bradtaniguchi: Don't use `let` if you can use `const`

src/utilities/formatter.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param {string} unformattedCode The unformatted code recieved from user's message
+ * @param {string} language Target language to format to
+ *
+ * @returns {string} The formatted code or a codeblock showing
+ * where syntax error occured
+ */
+
+module.exports = function formatter(unformattedCode, language) {
+  let options = {
```
bradtaniguchi: don't use `let` when you can use `const`

src/utilities/formatter.js
bradtaniguchi: Makes sense, and Ah good to know!

Put some of what you just told me as a comment above this line so whoever looks at this code down the line knows why its there.

Ryuno-Ki: Correct. We want to defer to Discord for syntax highlighting. Prettier is applying some ANSI styles as well (via Chalk).
I suggested three options:
- using a library, like stripAnsi (c.f. https://github.com/prettier/prettier/issues/2956#issuecomment-333597851 )
- use `--no-color` CLI argument (c.f. https://github.com/prettier/prettier/issues/4118#issue-303283804 )
- set `FORCE_COLOR=0` environment variable (c.f. https://github.com/prettier/prettier/issues/4118#issuecomment-371401955 )

src/utilities/process-reply.js
```diff
@@ -0,0 +1,84 @@
...
+  languageGuesses,
+  content
+}) {
+  const filter = (reaction) => {
+    return reactionEmojies.includes(reaction.emoji.name);
+  };
+
+  ref
+    .awaitReactions(filter, { max: 1, time: 60000, errors: ['time'] }) // code can be formatted multiple times by chnaging this line
+    .then((collected) => {
```
bradtaniguchi: could change to async/await

src/utilities/confused-message-generator.js
Ryuno-Ki: > Don't use a `.map` if you don't use the return value

In this case, `.forEach` would be more appropriate.
Alas, I second Brad.

src/commands/add-formatting.js
Twaha-Rahman: Thanks for the handy tips @bradtaniguchi !

src/commands/add-formatting.js
Twaha-Rahman: This command is triggered by any user's `bot`  emoji reaction to any message.

src/utilities/formatter.js
Twaha-Rahman: Done!

src/commands/add-formatting.js
Twaha-Rahman: Done!

src/utilities/confused-message-generator.js
Twaha-Rahman: Done!

src/commands/add-formatting.js
Twaha-Rahman: Changed code to use `async/await`...

src/utilities/process-reply.js
Twaha-Rahman: Done!

src/commands/add-formatting.js
Twaha-Rahman: The prefix doesn't really apply to this command.... infact, I don't know if we can call it a command at all since it listens for reaction rather than messages. Maybe move it to a separate directory all together?

bradtaniguchi: Yea I'd move this file somewhere else, since its not a command.

src/commands/add-formatting.js
```diff
@@ -0,0 +1,57 @@
...
+      reactionOptionsObj,
+      languageGuesses,
+      content
+    };
+
+    try {
+      const refToPromptMsg = await message.channel.send(confusedMessage);
+      processReply({ ...requiredInfoObj, ref: refToPromptMsg });
+    } catch (error) {
+      // maybe write an error log or something
```
bradtaniguchi: add `console.error`, rather get spammed then have the code suck up errors without letting us know.

src/index.js
```diff
@@ -19,6 +20,45 @@ const client = new Client();
...
+        try {
+          await reaction.fetch();
+        } catch (error) {
+          try {
+            reaction.message.channel.send(
+              'Something went wrong! Failed to format code :('
+            );
+          } catch (error) {
+            // maybe write an error log or something
+          }
```
bradtaniguchi: I don't think we the nested `try/catch`, just `console.error` the `error` return from the fetch call and send a message back to chat if possible.

If the bot can't send error messages to chat, then its fine if it explodes, as long as the system/service managing the bot notifies someone. For the official FCC version of the bot we probably wouldn't want to spam chat with error messages, but this version can. If we supress error messages via this nested try and catch **no one** will know what sort of errors are happening or why.

Assuming we host the bot on some Paas, or FCC uses some system like Docker/kubernetes, then the "host" system will manage critical errors like this, rather then the bot itself.

src/index.js
```diff
@@ -19,6 +20,45 @@ const client = new Client();
...
+
+          addFormatting.command(reaction.message);
+        } catch (error) {
+          try {
+            reaction.message.channel.send(
+              'Something went wrong! Failed to format code :('
+            );
+          } catch (error) {
+            // maybe write an error log or something
+          }
```
bradtaniguchi: Same as the other comment on nested try/catch blocks

src/utilities/process-reply.js
```diff
@@ -0,0 +1,88 @@
...
+        }
+        break;
+      }
+    }
+  } catch (error) {
+    try {
+      ref.channel.send('Something went wrong! Failed to format code :(');
+    } catch (error) {
+      // maybe write an error log or something
+    }
```
bradtaniguchi: Same as the others.

src/commands/add-formatting.js
Twaha-Rahman: Done...

src/index.js
Twaha-Rahman: Done...

src/index.js
Twaha-Rahman: Done...

src/utilities/process-reply.js
Twaha-Rahman: Done...

bradtaniguchi (commented):
So code LGTM, but on my end when testing the bot's format command using the `:robot:` emoji on the following code:
```js
console.log("Hello World!");
```

I get the following error from the discord api:
```
DiscordAPIError: Missing Permissions
    at RequestHandler.execute (/home/bradtaniguchi/Projects/discord-bot-test/node_modules/discord.js/src/rest/RequestHandler.js:170:25)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  method: 'delete',
  path: '/channels/709934973997678665/messages/712192004213440546/reactions',
  code: 50013,
  httpStatus: 403
}
```

The only `delete` we are performing has to deal with the emojis in the `src/index.js` file with `.removeAll()`. I guess the bot needs [specific permissions](https://discordjs.guide/popular-topics/permissions.html) for this to work.

Removing the `await reaction.message.reactions.removeAll();` line gets the feature to work (which ends up suggesting the languages `1c` and `abnf`, but that's another story that can be fixed later (#40).


There also seems to be an error returned to the client-side if they don't select one of the provided choices in the time frame provided by process reply, but this also can be fixed later.

--- 
**So whoever has a local run of this version of the bot working locally, please let me know what I have configured wrong on my bot with this current PR so we can document what we need to turn on to get the "removeAll" method working.** Otherwise, what is the impact of removing the `removeAll` emoji call?

Twaha-Rahman:
The bot requires requires `Manage Messages` permission to work. This is due to the bot's action of removing reactions from the message (This is done to prevent the bot from reacting multiple times to the same reaction). 

You'll have to give the bot `Manage Messages` permission either explicitly from permissions panel or by inviting it into the server using a link like this-

https://discordapp.com/oauth2/authorize?client_id=YOUR-CLIENT-ID&scope=bot&permissions=8192

(P.S. : Replace `YOUR-CLIENT-ID` with the bot's client id)

For more info go [here](https://discordjs.guide/preparations/adding-your-bot-to-servers.html#bot-invite-links).

bradtaniguchi:
> The bot requires requires `Manage Messages` permission to work. This is due to the bot's action of removing reactions from the message (This is done to prevent the bot from reacting multiple times to the same reaction).
> 
> You'll have to give the bot `Manage Messages` permission either explicitly from permissions panel or by inviting it into the server using a link like this-
> 
> https://discordapp.com/oauth2/authorize?client_id=YOUR-CLIENT-ID&scope=bot&permissions=8192
> 
> (P.S. : Replace `YOUR-CLIENT-ID` with the bot's client id)
> 
> For more info go [here](https://discordjs.guide/preparations/adding-your-bot-to-servers.html#bot-invite-links).

Yup looks to have been missing that one! I'll open another issue to update the README with those instructions so it can be part of setup instructions for this bot :)

Twaha-Rahman:
> Yup looks to have been missing that one! I'll open another issue to update the README with those instructions so it can be part of setup instructions for this bot :)

Yeah.... the README is getting outdated.... I think I'll get around to it after this PR...

bradtaniguchi (approved):
LGTM 👍 🚀 :shipit:

PR #29, opened 2020-05-16
Fix ESLint and Prettier rule clashing

### LINKED ISSUE #28: [Bug] Prettier's formatting clashing with ESlint
**Describe the bug**

Currently the Prettier extension formats the code according to its setting but ESLint doesn't allow some of those changes which gets linted by ESLint. This behavior can't be escaped as Prettier is currently being runned at commits, for which commits have linting issues which in tern fails in the Travis CI tests.

**To Reproduce**
1. Just write a `switch` statement commit it
2. ESLint will lint Prettier's formatting

**Expected behavior**

Prettier's formatting should remove ESLint formatting issues, not create them.

**Screenshots**

Capture

Twaha-Rahman:
This PR should fix #28

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,20 +2,19 @@ module.exports = {
   env: {
     commonjs: true,
     es6: true,
-    node: true,
+    node: true
   },
-  extends: 'eslint:recommended',
+  extends: 'plugin:prettier/recommended',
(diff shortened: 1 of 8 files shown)


bradtaniguchi (approved):
LGTM! 👍 🚀

PR #32, opened 2020-05-18
Use prefix from config and implement dynamic command handling

### LINKED ISSUE #11: Add support for configurable prefix
the current prefix is `!prefix` as a POC, we should look into changing this to a config option  with a sensible default.

cjcanlas01:
- Use prefix from get-config
- Implement dynamic command handling by setting all command files to a Discord Collection
- Remove `!` from prefix variable on the command file

Closes: #11

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -1,13 +1,16 @@
-const eightball = require('./eightball');
-const help = require('./help');
 const Discord = require('discord.js');
+const fs = require('fs');
 /**
  * Bootstraps all commands to the client.
  * @param {Object.client} client the discord client
  * @param {Object.config} config the application config
  */
 module.exports = function bootstrap({ client, config }) {
-  const commands = [eightball, help];
+  // Get all the command files from commands folder
+  const commands = fs
+    .readdirSync(__dirname)
+    .filter((file) => file.endsWith('.js'));
+  client.commands = new Discord.Collection();
 
   client.on('guildMemberAdd', (member) => {
     const welcomeEmbed = new Discord.MessageEmbed()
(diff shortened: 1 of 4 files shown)


src/commands/eightball.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  prefix: '!eightball',
+  prefix: 'eightball',
```
bradtaniguchi, line 2: The `!help` command also needs to be updated

**edit** with the changes merged in master, the `!stats` command also needs to be updated

src/commands/bootstrap.js
```diff
@@ -28,11 +31,36 @@ module.exports = function bootstrap({ client, config }) {
...
+  }
+
   client.on('message', (message) => {
-    for (let command of commands) {
-      if (message.content.startsWith(command.prefix)) {
-        command.command(message);
-        break;
+    if (message.content.startsWith(config.PREFIX)) {
+      // Get command after prefix
+      let commandArgument = message.content.split(' ')[1];
```
bradtaniguchi: change to `const` instead of `let`

src/commands/bootstrap.js
```diff
@@ -28,11 +31,36 @@ module.exports = function bootstrap({ client, config }) {
...
+
+      // Check if there are no commands sent
+      if (commandArgument == undefined) {
```
bradtaniguchi: Change to `!commandArgument` instead

bradtaniguchi (changes requested):

src/commands/eightball.js
naomi-lgbt: All that needs to change is the removal of the `!` for the prefix value, correct?

src/commands/bootstrap.js
cjcanlas01: Will do.

src/commands/eightball.js
cjcanlas01: Yes.

cjcanlas01:
@bradtaniguchi 

Pull request updated.

bradtaniguchi (approved):
LGTM 👍 🚀 :shipit:

PR #33, opened 2020-05-18
Make help command auto-updating

### LINKED ISSUE #22: Make the help command auto-updating.
Make the `help` command of the bot better, because currently, I assume whenever a new command is programmed, you manually have to add the name and description to the `help` command embed. Instead you can use the library to get all the command names with their description automatically and then iterate through them, appending fields to the embed. But an embed can only have 25 fields, so after each 25 index, you DM whoever invoked the command with the current batch. After that, the fields can be cleared and new command names/description can be added.

naomi-lgbt:
Help command will now do the following:

- [x] Automatically scan the files in the commands directory for commands.
- [x] Create embed with fields, each containing a command's prefix and description.
- [x] Skip any file in the commands directory that is missing either the prefix or description property. 
- [x] Trim the name and value of each embed field to ensure it does not exceed the limits. 

Closes #22

### DIFF
--- a/src/commands/help.js
+++ b/src/commands/help.js
@@ -1,5 +1,6 @@
 const Discord = require('discord.js');
 const fs = require('fs');
+const fsPromises = fs.promises;
 module.exports = {
   prefix: '!help',
   description: 'Get the commands currently available with this bot',
@@ -23,24 +24,24 @@ module.exports = {
         footer: { text: 'I am not affiliated with FreeCodeCamp in any way.' }
       };
 
-      fs.readdir(__dirname, (error, files) => {
-        if (error) {
-          console.error(error);
-        }
-        for (let i = 0; i < files.length; i++) {
-          let filename = files[i];
-          let lookup = require(`./${filename}`);
-          if (lookup.prefix == undefined || lookup.description == undefined) {
-            continue;
-          } else {
-            let fieldObj = {
-              name: lookup.prefix.substring(0, 255),
-              value: lookup.description.substring(0, 1023)
-            };
-            helpEmbed.fields.push(fieldObj);
+      fsPromises
+        .readdir(__dirname)
+        .then((result) => {
+          for (let i = 0; i < result.length; i++) {
+            const filename = result[i];
+            const lookup = require(`./${filename}`);
+            if (!lookup.prefix || !lookup.description) {
+              continue;
+            } else {
+              const fieldObj = {
+                name: lookup.prefix.substring(0, 255),
+                value: lookup.description.substring(0, 1023)
+              };
+              helpEmbed.fields.push(fieldObj);
+            }
           }
-        }
-      });
+        })
+        .catch((error) => console.log(error));
 
       message.author.send({ embed: helpEmbed });
     } catch (error) {
--- a/src/commands/help.js
+++ b/src/commands/help.js
@@ -1,46 +1,49 @@
-const Discord = require('discord.js');
+const Discord = require("discord.js");
+const fs = require("fs");
 module.exports = {
-  prefix: '!help',
+  prefix: "!help",
+  description: "Get the commands currently available with this bot",
   /**
    * @name help
    * Displays currently available commands.
    *
    * @param {Discord.Message} message the message provided
-   *Use this format to add new commands:
-   *{name: "Command Name", value: "Command Description"},
+   * Now automatically adds new commands - make sure the .js file has a prefix and description,
    */
 
   command: async function help(message) {
     //console.log("help", message);
     try {
-      const helpEmbed = new Discord.MessageEmbed()
-        .setColor('#0099ff')
-        .setTitle('Bot Information')
-        .setDescription(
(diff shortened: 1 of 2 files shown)


Twaha-Rahman:
Looks like an interesting addition. Well done @nhcarrigan !

naomi-lgbt:
Thanks! Not sure how well it will mesh with #32 but we can tweak it as needed I'm sure.

src/commands/help.js
```diff
@@ -1,44 +1,48 @@
...
+      };
 
-          //for spacing, leave this one at bottom.
-          { name: '\u200B', value: '\u200B' }
-        )
-        .setFooter('I am not affiliated with FreeCodeCamp in any way.');
+      fs.readdir(__dirname, (error, files) => {
+        if (error) {
+          console.error(error);
+        }
```
bradtaniguchi: Change this to the `promise` version:
https://nodejs.org/dist/latest-v10.x/docs/api/fs.html

src/commands/help.js
```diff
@@ -1,44 +1,48 @@
...
+        for (let i = 0; i < files.length; i++) {
+          let filename = files[i];
+          let lookup = require(`./${filename}`);
+          if (lookup.prefix == undefined || lookup.description == undefined) {
```
bradtaniguchi: Don't use `==`, use `===`, or just check for "falsy":
```js
if (!lookup.prefix || !lookup.description) {
```

src/commands/help.js
```diff
@@ -1,44 +1,48 @@
...
-          { name: '\u200B', value: '\u200B' }
-        )
-        .setFooter('I am not affiliated with FreeCodeCamp in any way.');
+      fs.readdir(__dirname, (error, files) => {
+        if (error) {
+          console.error(error);
+        }
+        for (let i = 0; i < files.length; i++) {
+          let filename = files[i];
+          let lookup = require(`./${filename}`);
```
bradtaniguchi: Don't use `let` if you can use `const`. (this should be made a linter rule, but that can be done later)

bradtaniguchi (changes requested):

src/commands/help.js
naomi-lgbt: Done :)

src/commands/help.js
naomi-lgbt: Wasn't sure if I could use const within the loop. Corrected.

src/commands/help.js
naomi-lgbt: Believe this is implemented correctly now.

src/commands/help.js
```diff
@@ -1,44 +1,49 @@
...
+        footer: { text: 'I am not affiliated with FreeCodeCamp in any way.' }
+      };
 
-          //for spacing, leave this one at bottom.
-          { name: '\u200B', value: '\u200B' }
-        )
-        .setFooter('I am not affiliated with FreeCodeCamp in any way.');
+      fsPromises
+        .readdir(__dirname)
+        .then((result) => {
```
bradtaniguchi: Change to `async/await`, we are trying not to use `.then` and `.catch` when possible.
Also change `result` to `results`, as its an array, and thus should be named plural.

src/commands/help.js
```diff
@@ -1,44 +1,49 @@
...
+            if (!lookup.prefix || !lookup.description) {
+              continue;
+            } else {
+              const fieldObj = {
+                name: lookup.prefix.substring(0, 255),
+                value: lookup.description.substring(0, 1023)
+              };
+              helpEmbed.fields.push(fieldObj);
+            }
+          }
```
bradtaniguchi: I'm not a fan of the classic for loop, `.forEach`, or [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is less to type, and is less likely to get screwed up.

bradtaniguchi (changes requested):

src/commands/help.js
naomi-lgbt: Changed to .forEach

src/commands/help.js
naomi-lgbt: Think I got what you were saying.

src/commands/help.js
```diff
@@ -1,46 +1,42 @@
...
+        if (lookup.prefix && lookup.description) {
+          const fieldObj = {
+            name: lookup.prefix.substring(0, 255),
+            value: lookup.description.substring(0, 1023)
+          };
+          helpEmbed.fields.push(fieldObj);
+        }
+      });
+      await message.author.send({ embed: helpEmbed });
+    });
```
bradtaniguchi: Update to:
```js
const results = await fsPromises.readdir(__dirname);
console.log(results);
//... rest of code
```

naomi-lgbt: Should be good to go now!

bradtaniguchi (approved):
LGTM 👍 🚀 :shipit:

Issue #40, opened 2020-05-19
Decrease supported languages for Highlightjs


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
After merging #27, the bot will get "confused" between potential language guess more often than not. This is due to the **huge** amount of language support the library has

**Describe the solution you'd like**
After merging #27, the bot will provide 2 guesses for a given language if it can't decide. We can decrease the collision amount by providing only a set list of language options. The list of supported languages should be more or less the same as those that can be formatted by prettier and maybe a few more that are commonly asked for within FCC.

The most common use-case languages will probably be html, css, JS and Python (which isn't supported by Prettier, but is usually readable automatically). We can support a few other ones that prettier supports that I haven't listed.

So we will update the [hljs language set](https://highlightjs.readthedocs.io/en/latest/api.html#highlightauto-code-languagesubset) to be only from a limited set of commonly used languages within the FCC community.

**Additional context**
This should decrease the available "features" the syntax highlighting can recognize, but make the bot easier to use for most of its common use-cases.

naomi-lgbt:
According to the highlight.js docs, [these](https://highlightjs.org/static/demo/) are the common languages - might be worth considering this list? 39 is a lot fewer than 189.

Twaha-Rahman:
I'm working on it.... @bradtaniguchi feel free to assign it to me..

PR #42, opened 2020-05-19
Limit Highlight.js language detection

Twaha-Rahman:
This PR addresses #40

### DIFF
--- a/src/utilities/process-reply.js
+++ b/src/utilities/process-reply.js
@@ -38,7 +38,7 @@ module.exports = async function processReply({
         if (supportedLanguage) {
           const formattedCode = formatter(content, supportedLanguage);
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[0], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
           );
         } else {
           ref.channel.send(
@@ -53,7 +53,7 @@ module.exports = async function processReply({
           const formattedCode = formatter(content, supportedLanguage);
 
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[1], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
           );
         } else {
           ref.channel.send(
@@ -68,7 +68,7 @@ module.exports = async function processReply({
         if (supportedLanguage) {
           const formattedCode = formatter(content, supportedLanguage);
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[2], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
           );
         } else {
           ref.channel.send(
@@ -79,7 +79,7 @@ module.exports = async function processReply({
       }
     }
   } catch (error) {
-    ref.channel.send('Something went wrong! Failed to format code :(');
+    ref.channel.send('Message timeout!');
     console.error(error);
   }
 };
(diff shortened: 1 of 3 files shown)


src/utilities/process-reply.js
```diff
@@ -38,7 +38,7 @@ module.exports = async function processReply({
         if (supportedLanguage) {
           const formattedCode = formatter(content, supportedLanguage);
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[0], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
```
bradtaniguchi, line 41: If we are just using `supportedLanguage` now, what about the `languageGuesses` param? Does the bot no longer ever give multiple language suggestions?

src/utilities/process-reply.js
```diff
@@ -53,7 +53,7 @@ module.exports = async function processReply({
           const formattedCode = formatter(content, supportedLanguage);
 
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[1], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
```
bradtaniguchi, line 56: This no longer seems possible, if we are in the `else` case but the `if` case is if there is a `supportedLanguage` in the first place?

src/utilities/process-reply.js
```diff
@@ -68,7 +68,7 @@ module.exports = async function processReply({
         if (supportedLanguage) {
           const formattedCode = formatter(content, supportedLanguage);
           ref.channel.send(
-            formatMessageWithCodeblock(languageGuesses[2], formattedCode)
+            formatMessageWithCodeblock(supportedLanguage, formattedCode)
```
bradtaniguchi: Isn't this also impossible?

bradtaniguchi (changes requested):

src/utilities/process-reply.js
Twaha-Rahman: This was the third option when `language-classifier` was used....yes this code is unreachable... I'll look into it in more details after a few hours...

Twaha-Rahman:
@bradtaniguchi The bot gives multiple suggestions just like before..... I have only changed the `languageGuesses[0]` and `languageGuesses[1]` with `supportedLanguage` (If the language can be formatted with `Prettier`) . This change is made to change some detected languages name when we return the message to Discord, because even though `Prettier` supports the lang, Discord doesn't support it. For example- `SCSS` gets's detected and formatted with `Prettier` but Discord doesn't support syntax highlighting for it..... so we change `SCSS` to `CSS` when we format the message for Discord.

In this line `const supportedLanguage = isSuppotedByPrettier(languageGuesses[0]);` the `isSuppotedByPrettier` function checks if the lang is supported by `Prettier`.... if it isn't, the function returns false... that's why we check if the variable `supportedLanguage ` is false...

src/utilities/process-reply.js
Twaha-Rahman: I've removed the third `case` in the `switch` statement as it is unreachable after removing `language-classifier`.

src/utilities/process-reply.js
Twaha-Rahman: I've explained this details in the message below...

bradtaniguchi (approved):
Just tested this in my bot room again and it looked to work with a I posted earlier and it worked. I also accidentally tested this on an already formatted message provided from the bot, and the formatted code ended up really funky, but I don't think that is that big of a deal.

LGTM :+1:

PR #44, opened 2020-05-19
Add thank command

### LINKED ISSUE #30: Add "Thanks" support back
The previous gitter bot had a "brownie points" feature, which **will not** be replicated due to spam and "gaming the system" issues. However, having a "hollow" version I think would be acceptable. 

Anytime someone thanks someone else, the bot will "give brownie points" to the other user. This wont actually increase anyone's points or anything, but will bring recognition to those that help others, which was the main goal of the originals.

**Additional context**
~~TDB (I want to add a reference to the originals text/emoji/setup, we will use the same one for this iteration)~~

**edit** finally, here is the original source code for the original message format:
https://github.com/freeCodeCamp/camper-gitter-bot/blob/529cc6e989f2f9661e3dd65d927727c6489c8249/lib/bot/cmds/thanks.js#L49

(I think this is already know, but adding it anyways)

naomi-lgbt:
Thank command will do the following: 
- [x] Read each message for the word "thanks"
- [x] Check if the following word is a **tagged** user.
   - [x] If the following word is a tagged user, send a message that the author sends that user brownie 
points. Both author and user are tagged.
   - [x] If the tagged user is the author, send a message that the author cannot thank itself.
- [x] If the following word is **not** a tagged, user, do nothing.
- [x] Read the entire message and process each "thanks" independently.

Upon merge, this will close #30

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -63,7 +63,7 @@ module.exports = function bootstrap({ client, config }) {
         message.reply('there was an error trying to execute that command!');
       }
     }
-        //thank command
+    //thank command
     if (message.content.toLowerCase().includes('thanks')) {
       const messageArr = message.content.split(' ');
       for (let i = 0; i < messageArr.length; i++) {
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -63,5 +63,34 @@ module.exports = function bootstrap({ client, config }) {
         message.reply('there was an error trying to execute that command!');
       }
     }
+        //thank command
+    if (message.content.toLowerCase().includes('thanks')) {
+      const messageArr = message.content.split(' ');
+      for (let i = 0; i < messageArr.length; i++) {
+        if (messageArr[i] === 'thanks') {
+          const args = messageArr[i + 1];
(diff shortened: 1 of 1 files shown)


bradtaniguchi (changes requested):
So this feature is tricky in that there are multiple ways to thank people:
1. `Thanks @bradtaniguchi`
2. `Thank you @bradtaniguchi `
3. `@bradtaniguchi thanks`
4. `thanks @bradtaniguchi`
5. `thanks @bradtaniguchi @bradtaniguchi2`
(and others)

I should of mentioned this in the issue, to give some better examples of when this should happen or not.

I think this is a situation where jest should come in, as I was able to crash the server the current code with just:
```
@bradtaniguchi thanks
```

and most of the above cases don't currently work.
 
I'll refactor some of this, move the function to a separate file and setup a few test-cases, with an example one already setup (the current setup that "passes") so we can easily test this against any future message patterns. 

This way we can take a test driven development approach with this feature.

bradtaniguchi:
Closing in favor of this commit, used your changes as a basis though!
bb66c2cea89e64c0509f71a7f5fa139cc4f9a668

PR #45, opened 2020-05-19
Update eslint rules to prevent let when const can be used

### LINKED ISSUE #38: Update eslint rules to prevent let, when const will do
We can write `let` instead of `const` currently, but this isn't neccessary.

Update eslint rules to prevent using `let` if a `const` will do.

**Additional context**
Docs on possible solution:
https://eslint.org/docs/rules/prefer-const

Twaha-Rahman:
This PR closes #38

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/src/tests
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-/src/tests
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,8 @@ module.exports = {
     es6: true,
     node: true
   },
-  extends: 'plugin:prettier/recommended',
+  plugins: ['prettier'],
+  extends: ['eslint:recommended', 'plugin:prettier/recommended'],
   globals: {
     Atomics: 'readonly',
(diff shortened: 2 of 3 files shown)


.eslintignore
```diff
@@ -0,0 +1 @@
+/src/tests
```
bradtaniguchi: tests should still be covered though right?

Twaha-Rahman: ESLint goes crazy when it sees we are using Jest functions without importing it... that's why I had to ignore it

Ryuno-Ki: Was there no way to declare it as a global?

bradtaniguchi: I looked into this and found that we can remove the `.eslintignore` file, so we get linting in our tests. 
To add eslint support for jest all we need to do is update the `eslintrc.js` file's `env` property like so:
```js
module.exports = {
env: {
  //..
  jest: true
},
//..
```

[ref](https://stackoverflow.com/questions/31629389/how-to-use-eslint-with-jest)

Twaha-Rahman: Ok... updating the PR!

Twaha-Rahman:
@bradtaniguchi Done! I think we're ready to go!

PR #48, opened 2020-05-31
bug fix

### LINKED ISSUE #47: [Bug] Server crash when user leaves
**Describe the bug**
When a user leaves, the whole server crashes and needs to be started again

**To Reproduce**
1. Leave the server

**Expected behavior**
Send a leave message to some channel

**Screenshots**
bilde

naomi-lgbt:
Closes #47

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -28,14 +28,14 @@ module.exports = function bootstrap({ client, config }) {
   });
 
   client.on('guildMemberRemove', function (member) {
-    const goodbye = member.guild.channels.cache.find(
+    const goodbyeChannel = member.guild.channels.cache.find(
       (channel) => channel.name == 'introductions'
     );
-    if (goodbye == undefined) {
-      console.log('Channel not found.');
+    if (!goodbyeChannel) {
+      console.error('goodbye channel not found.');
       return;
     } else {
-      goodbye.send(`** ${member.user} has left us! :( **`);
+      goodbyeChannel.send(`** ${member.user} has left us! :( **`);
     }
   });
   for (const file of commands) {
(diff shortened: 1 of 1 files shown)


src/commands/bootstrap.js
```diff
@@ -29,7 +29,7 @@ module.exports = function bootstrap({ client, config }) {
 
   client.on('guildMemberRemove', (member) => {
     member.guild.channels.cache
-      .find((ch) => ch.name === 'introduction')
+      .find((ch) => ch.name === 'introductions')
```
bradtaniguchi: This might be more useful as a config setting. (this can be done in another issue/pr)

also, to prevent the server from exploding check that `.find` returns something before calling `.send`. If it **doesn't** find the channel specified log an error about how the "remove channel" wasn't found or something.

bradtaniguchi (changes requested):

src/commands/bootstrap.js
naomi-lgbt: Should be corrected now.

naomi-lgbt:
Okay, I tested this code on my bot and it logs the "Channel not found" message and does not crash the server.

src/commands/bootstrap.js
```diff
@@ -27,12 +27,17 @@ module.exports = function bootstrap({ client, config }) {
...
+  client.on('guildMemberRemove', function(member) {
+  const goodbye = member.guild.channels.cache.find(
+    channel => channel.name == 'introductions'
+  );
+  if (goodbye == undefined) {
+    console.log('Channel not found.');
+    return;
+  } else {
+    goodbye.send(`** ${member.user} has left us! :( **`);
+  }
```
bradtaniguchi: Change to just:
```
if (!goodebyeChannel) {
  console.error('goodbye channel not found');
  return;
}
goodebyeChannel.send(/**/);
```

naomi-lgbt: Better? :)

bradtaniguchi (approved):

PR #53, opened 2020-06-09
Add feature flag for deleted messages logging

### LINKED ISSUE #52: Add feature flag for moderation-activity room name
Right now the `moderation-activity` channel is hardcoded. This should be changed to an environment variable so we can activate/de-activate this feature along with change the room.

This will also prevent the bot from replying to the existing channel if the room isn't found.

Update the `get-config` function to get an environment variable, and set the default.

naomi-lgbt:
Okay, I don't have a local version of your bot to test this on. But I followed the example for the depart messages (except the secret .env file doesn't exist so I can't set that up.) and I believe I've configured the logging to call the variable correctly.
Also changed it to log nothing (just throw a console message) if the channel doesn't exist.

Closes #52

### DIFF
--- a/src/config/get-config.js
+++ b/src/config/get-config.js
@@ -42,7 +42,7 @@ module.exports = function getConfig() {
      * If not given, or if the channel is not found
      * the log will not be generated.
      */
-    LOG_MSG_CHANNEL: process.env.LOG_MSG_CHANNEL === 'true',
+    LOG_MSG_CHANNEL: process.env.LOG_MSG_CHANNEL || 'moderation-activity',
     /**
      * The port to "listen" for liveness checks.
      * Defaults to 8080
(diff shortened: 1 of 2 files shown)


src/config/get-config.js
```diff
@@ -37,6 +37,12 @@ module.exports = function getConfig() {
      * the message will not be provided.
      */
     LEAVE_MSG_CHANNEL: process.env.LEAVE_MSG_CHANNEL === 'true',
+    /**
+     * The name of the channel to log deleted messages to.
+     * If not given, or if the channel is not found
+     * the log will not be generated.
+     */
+    LOG_MSG_CHANNEL: process.env.LOG_MSG_CHANNEL === 'true',
```
bradtaniguchi: Change to:
```
 LOG_MSG_CHANNEL: process.env.LOG_MSG_CHANNEL || 'moderation-activity',
```

naomi-lgbt: Done. :)

bradtaniguchi (approved):

PR #56, opened 2020-06-16
Add attachment filtering

naomi-lgbt:
Removes any message containing an attachment that is not an image or video, to prevent sending harmful code files.

### DIFF
--- a/src/commands/bootstrap.js
+++ b/src/commands/bootstrap.js
@@ -121,7 +121,10 @@ module.exports = function bootstrap({ client, config }) {
   });
 
   client.on('message', (message) => {
-    if (message.attachments.array().length > 0 && !message.attachments.array()[0].height) {
+    if (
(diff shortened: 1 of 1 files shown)


src/commands/bootstrap.js
```diff
@@ -121,6 +121,14 @@ module.exports = function bootstrap({ client, config }) {
   });
 
   client.on('message', (message) => {
+    if (message.attachments.array().length > 0) {
+      if (!message.attachments.array()[0].height) {
```
bradtaniguchi: combine the 2 if statements into one with `&&`
```
if (message.attachments.array().length && !message.attachments.array()[0].height) {
  //...
}
```

bradtaniguchi (changes requested):

bradtaniguchi (approved):

PR #57, opened 2020-06-18
Chore/55/migrate to ts

### LINKED ISSUE #55: Refactor project to TypeScript
Currently the project is written in JS for lower barrier of entry, however discord.js already provides first party TS support via providing their own types. 

We should migrate to TS which will make it slightly harder to develop, **but** make it easier to leverage discord features thru the provided types. 

We **don't** need TS, but I believe having it will prevent bugs down the line, and increase overall code quality.

**Additional context**
Due to how the code is currently written, we should be able to make minimal changes to the existing code to get everything to work.

bradtaniguchi:
- Update all code to TS

Closes #55 

**note** this is a draft as I need to update the github action deployment to take into account the new TS flow


src/config/get-config.ts
```diff
@@ -0,0 +1,63 @@
...
+   * By default this feature is OFF to suppress the amount of
+   * logs being printed.
+   */
+  VERBOSE: boolean;
+  /**
+   * If the bot should send the FCC code of conduct via a DM
+   * to any new user who joins the server.
+   * By default this feature is OFF
+   */
+  WELCOME_DM: boolean;
```
naomi-lgbt, line 33: Quick question. This is typed as a boolean here, but compared to a string in the return. 
So my `.env` had to have it as a string and wouldn't register if I used the boolean.

src/commands/bootstrap.ts
```diff
@@ -78,24 +84,22 @@ module.exports = function bootstrap({ client, config }) {
...
-    // Set a new command file in the Discord Collection
-    // With the key as the command prefix and the value as the exported command function
-    client.commands.set(command.prefix, command);
-  }
-
   //deleted message logging
   client.on('messageDelete', function (message) {
     //change channel name to match server configuration
-    const logChannel = message.guild.channels.cache.find(
+    const logChannel = message.guild?.channels.cache.find(
```
naomi-lgbt: Question mark here is breaking the deleted message logging function. 😶

bradtaniguchi: This feature was working on my end, let me debug to see if I can get it to break.

What "broke" when this was ran? Did you get any errors or anything?

I'm adding an extra check around this feature to check `config.LOG_MSG_CHANNEL` is provided at all before adding this callback, so if its not available this code wont ever ran at all.

src/config/get-config.ts
bradtaniguchi: All environment variables provided from `process.env` should be strings or undefined. 
So if your using `.env` then either of the following should end up as "true":
```
WELCOME_DM=true
```
or
```
WELCOME_DM='true'
```

src/index.ts
```diff
@@ -1,8 +1,8 @@
-const { Client } = require('discord.js');
-const express = require('express');
-const getConfig = require('./config/get-config');
-const validateConfig = require('./config/validate-config');
-const bootstrap = require('./commands/bootstrap');
+import { Client } from 'discord.js';
+import express from 'express';
```
Twaha-Rahman, line 2:
```suggestion
import express, { Request, Response } from 'express';
```

Then, on line 29 -

```js
    expressApp.get('/status', (_: Request, res: Response) =>
```

Twaha-Rahman (changes requested):
I think we should be as explicit with our types as possible.

src/index.ts
bradtaniguchi: Explicit types are not necessary, as the `expressApp` will infer types, so the arguments are typed without us passing anything.

I'm not of fan of adding types for the sake of adding types as it increases code bloat and decreasing developer productivity, especially for those that aren't used to TS.

src/commands/bootstrap.ts
naomi-lgbt: Adding the `LOG_MSG_CHANNEL` to the `.env` didn't allow the command to work until I removed the `?` and recompiled.

I'll try testing it again to see if it was an error on my part.

src/config/get-config.ts
naomi-lgbt: `WELCOME_DM='true'` worked for me, but `WELCOME_DM=true` did not. 
I'll try testing it again to see if it was an error on my end.

bradtaniguchi: > WELCOME_DM='true' worked for me, but WELCOME_DM=true did not.
I'll try testing it again to see if it was an error on my end.

I'll recheck with `WELCOME_DM=true`

src/commands/bootstrap.ts
bradtaniguchi: I'm on node version (node -v) `12.17.0` and have `tsc` version (npx tsc -v) of `3.9.5`.

@nhcarrigan Do you see an error during build (its a TS error) or when that specific code is ran (its a JS error)?

naomi-lgbt: I recompiled with the `?` present and now it seems to be functioning fine. I must have made an error last time. 

Running the `npm run build` only throws the `cannot find type definition file for 'jest'` error. I'm running node version `14.2.0` and apparently only `3.9.3` of `tsc`.

src/config/get-config.ts
naomi-lgbt: This seems to be working now - I must have royally messed up my initial testing 😬

src/index.ts
Twaha-Rahman: Ok... got it.

bradtaniguchi:
@nhcarrigan @Twaha-Rahman If you can, can you run the bot locally and just test out mainly the `close` and `suspend` commands which were newly added, and "converted" to TS. Everything else should be the same as the last review.

There are still issues related to the bot being added to the `suspend` user room, but that will be fixed after this PR is merged.
Let me know if you notice anything with the behavior or code from the migration. The logic should stay the same, but I mainly just fixed type issues where I saw them :D

PR #64, opened 2020-06-19
Enchancement/59/Update stats command to include more information about the bot.

JoshQuaintance:
Added Uptime on the stats command, showing the uptime for the bot.
Added Online Time on the stats command, showing what time the bot started.
Changed the command calling to pass in the Client variable to get Uptime.

### DIFF
--- a/src/commands/stats.js
+++ b/src/commands/stats.js
@@ -6,8 +6,44 @@ module.exports = {
    * Displays the server stats.
    * @param {Discord.Message} message the message provided by discord
    */
-  command: function stats(message) {
+  command: function stats(message, client) {
     try {
+      let totalSeconds = client.uptime / 1000;
+      const days =
+        Math.floor(totalSeconds / 86400) <= 0
+          ? ''
+          : Math.floor(totalSeconds / 86400) < 10
+          ? `0${Math.floor(totalSeconds / 86400)} days`
+          : `${Math.floor(totalSeconds / 86400)} days`;
+
+      const hours =
+        Math.floor(totalSeconds / 3600) <= 0
+          ? ''
+          : Math.floor(totalSeconds / 3600) < 10
+          ? `0${Math.floor(totalSeconds / 3600)} hours`
+          : `${Math.floor(totalSeconds / 3600)} hours`;
+
+      totalSeconds %= 3600;
+      const minutes =
+        Math.floor(totalSeconds / 60) <= 0
+          ? ''
+          : Math.floor(totalSeconds / 60) < 10
+          ? `0${Math.floor(totalSeconds / 60)}  minutes`
+          : `${Math.floor(totalSeconds / 60)} minutes`;
+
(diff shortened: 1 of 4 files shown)


src/commands/stats.js
```diff
@@ -6,8 +6,44 @@ module.exports = {
...
+          ? `0${Math.floor(totalSeconds % 60)} seconds`
+          : `${Math.floor(totalSeconds % 60)} seconds`;
+
+      const date = new Date();
+      const hour =
+        date.getHours() < 10 ? `0${date.getHours()}` : date.getHours();
+      const mins =
+        date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
+      const tz = /\((.*)\)/.exec(new Date().toString())[1];
+      const am_pm = hour >= 12 ? 'PM' : 'AM';
```
bradtaniguchi: Move the 2 distinct code blocks to their own functions that return the strings provided to the `statsEmbed` block.

bradtaniguchi (commented):
How do you feel about the custom date formatting you have? 

I'm the type of developer where I see a date calculation problem and walk straight to a lib solution (lately its been [luxon](https://moment.github.io/luxon/)) 

If your even slightly unsure about the code you have right now we can go use something like luxon instead and not worry about this code, nor worry about maintaining it if there are bugs/improvements down the line.

src/commands/stats.js
JoshQuaintance: I'll do this tomorrow. But I'll look into luxon and maybe even use it.

JoshQuaintance: Moved code blocks into each corresponding function that will return string that contains each corresponding data.

JoshQuaintance:
The `getWakeTime()` function doesn't work to it's full potential yet, it needs a JSON file to store it's data. I don't know if you want me to add a JSON file or something else, just clarifying.

bradtaniguchi:
> I'll do this tomorrow. But I'll look into luxon and maybe even use it.

I'll leave this PR open until you review Luxon, which could be used for both situations.

I'd also recommend renaming the two as `uptime` and `wakeTime` sound too similar, but this can be done later if you do use Luxon. 

> The getWakeTime() function doesn't work to it's full potential yet, it needs a JSON file to store it's data. I don't know if you want me to add a JSON file or something else, just clarifying.

If we need to have a JSON file with a date to represent "last deployed", then just return any date from this function, as it wont need to do any calculations, it will just show the date saved in the JSON file the bot is deployed with. (this can be done in another PR where we actually have the JSON file)

JoshQuaintance:
Made the function into it's own file so it can be called in the main file. On the bot start, it will generate and return the date, then it can be pushed into the JSON file.

bradtaniguchi (approved):
LGTM, the `botOnlineAt` variable might be saved differently in the future but should be fine enough for now.

PR #68, opened 2020-06-19
Fix/help

naomi-lgbt:
Because the bot sends the help message as a DM, it remains silent in the channel and appears unresponsive to other users. I've added a short confirmation to be sent to the command channel to show that the bot has received the command.

### DIFF
--- a/src/commands/help.js
+++ b/src/commands/help.js
@@ -37,5 +37,6 @@ module.exports = {
       }
     });
     await message.author.send({ embed: helpEmbed });
+    await message.channel.send('Help message sent!');
   }
 };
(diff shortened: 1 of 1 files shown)


src/commands/help.js
```diff
@@ -37,5 +37,6 @@ module.exports = {
       }
     });
     await message.author.send({ embed: helpEmbed });
+    await message.channel.send('Help message sent!');
```
bradtaniguchi: I don't think we need the 2nd await here

naomi-lgbt: You are correct - tested without and it still functioned as intended.

Issue #69, opened 2020-06-19
Add "suspend" action for moderators


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
Currently FCC mods handle suspended users via DMs, this needs to change as Discord DMs are easily disabled, making moderating users hard.

**Describe the solution you'd like**
- [ ] Moderator does command: `!suspend <user> "<reason>" "<evidence>"`
  - [ ] this action should be logged into the `LOG_MSG_CHANNEL`  automatically
- [ ] Bot gives the user the Suspended role
- [ ] Bot writes the report and puts it into #admin
- [ ] Bot creates a private channel only accessible by the suspended user and staff
- [ ] (optionally) Bot DM's the user about being Suspended and that a private channel has been created for them
- [ ] After the situation has been dealt with: !close to close that channel
Moderators can remove the Suspended role and / or ban the user depending on the outcome themselves. The bot doesn't have to get involved with that

**Additional context**
@nhcarrigan said he already has most of these features figured out on his own bot

naomi-lgbt:
I've got the following features on my own bot currently:
- [x] Moderator does command
- [x] Bot assigns role
- [x] Bot sends embed report to `LOG_MSG_CHANNEL`

Will start working on migrating those features to this codebase and adding new features as requested.

PR #70, opened 2020-06-19
Feature/69/add suspend command

naomi-lgbt:
* [X]  Moderator does command: `!suspend <user> "<reason>" "<evidence>"`
    * [X]  this action should be logged into the `LOG_MSG_CHANNEL`  automatically
 * [X]  Bot gives the user the Suspended role
 * [X]  Bot writes the report and puts it into #admin
 * [X]  Bot creates a private channel only accessible by the suspended user and staff
 * [X]  (optionally) Bot DM's the user about being Suspended and that a private channel has been created for them
 * [X]  After the situation has been dealt with: !close to close that channel
  Moderators can remove the Suspended role and / or ban the user depending on the outcome themselves. The bot doesn't have to get involved with that
---
Okay, I've got the code performing the above functions now. The `close` command will be in a separate PR, if we choose to use that feature. 
I can add the optional DM if desired.

Connected to #69


src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+const config = getConfig();
+module.exports = {
+  prefix: 'suspend',
+  description: 'Suspends a user for the given reason',
+  command: async function suspend(message) {
+    //check for appropriate permissions
+    if (message.member.hasPermission('KICK_MEMBERS') == false) {
+      message.channel.send(
+        `ERROR 401: ${message.author}, missing permissions.`
+      );
```
bradtaniguchi: Should we return errors back to the client like this into chat? Seems like spam. We should definitely log these errors via `console.error` or `console.log` if there is something wrong.

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
+const Discord = require('discord.js');
+const getConfig = require('../config/get-config.js');
+const config = getConfig();
+module.exports = {
+  prefix: 'suspend',
+  description: 'Suspends a user for the given reason',
+  command: async function suspend(message) {
+    //check for appropriate permissions
+    if (message.member.hasPermission('KICK_MEMBERS') == false) {
```
bradtaniguchi: Change conditional check to:
```
if (!message.member.hasPermission('KICK_MEMBERS)) {
```

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+      message.channel.send(
+        `ERROR 401: ${message.author}, missing permissions.`
+      );
+      return;
+    }
+    const mod = message.author;
+    const msgArguments = message.content.split(' ');
+    const user = message.mentions.members.first();
+    //check for valid user tag
+    if (user == undefined) {
```
bradtaniguchi: change to: 
```
if (!user) {
```

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+      message.channel.send(`ERROR 400: ${mod}, invalid user tag.`);
+      return;
+    }
+    //cannot target self
+    if (user === mod) {
+      message.channel.send(`ERROR 400: ${mod}, cannot target self.`);
+      return;
+    }
+    const reasonArg = msgArguments.slice(3, msgArguments.length);
+    let reason = reasonArg.join(' ');
```
bradtaniguchi: Change to just 
```
const reason = reasonArg.join(' ') || 'ERROR 404: No reason provided.'
```

If this is an error shouldn't we exit?

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+    );
+    if (modChannel) {
+      modChannel.send(restrictEmbed);
+    }
+    if (!modChannel) {
+      message.channel.send('ERROR 404: missing log channel');
+    }
+    //assign roles
+    user.roles.remove(user.roles).catch((e) => console.log(e));
+    user.roles.add(suspend).catch((e) => console.log(e));
```
bradtaniguchi: Shouldn't these fail the process and not fail silently? Also in a `catch` be sure to log errors as `console.error`

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+      .setFooter('Please remember to follow our rules!');
+    const modChannel = message.guild.channels.cache.find(
+      (channel) => channel.name === config.LOG_MSG_CHANNEL
+    );
+    if (modChannel) {
+      modChannel.send(restrictEmbed);
+    }
+    if (!modChannel) {
+      message.channel.send('ERROR 404: missing log channel');
+    }
```
bradtaniguchi: If no mod channel is found, shouldn't this fail early?

It might make sense to check for the 2 channels before we do anything else, as without these two channels being configured this feature wont work.

src/commands/suspend.js
```diff
@@ -0,0 +1,90 @@
...
+    //assign roles
+    user.roles.remove(user.roles).catch((e) => console.log(e));
+    user.roles.add(suspend).catch((e) => console.log(e));
+    //create suspend channel
+    const channelName = `suspended-${user.user.username}`;
+    const suspendCategory = config.SUSPEND_CATEGORY;
+    const category = await message.guild.channels.cache.find(
```
bradtaniguchi: shouldn't this error out if the category is missing?

src/config/get-config.js
```diff
@@ -43,6 +43,18 @@ module.exports = function getConfig() {
...
+     * If not given, or if the role is not found
+     * the command will not function
+     */
+    SUSPEND_ROLE: process.env.SUSPEND_ROLE || `null`,
+    /**
+     * The name of the category to create new
+     * suspended user channels in.
+     * If not given, channel will not be created.
+     */
+    SUSPEND_CATEGORY: process.env.SUSPEND_CATEGORY || `null`,
```
bradtaniguchi: these can default to nothing rather than a string of `null` which is **nasty** as `null` is falsy, but a non empty string is **truthy**.

bradtaniguchi (changes requested):
This command function probably will need to be broken down into a few different functions as its getting rather long. But first to refactor with the general following:

- add console.log/console.errors
- probably remove all the response message "ERRORS" to prevent chat spam from non mod users/ bad configs on our part. 
- any "null" expression checks in `if` statements can be simplified to negation checks.
also add an early escape from the command function if either environment variable isn't given, as the feature shouldn't work work without these 2 specified.

src/commands/suspend.js
naomi-lgbt: The command should still execute even if the user fails to provide a reason, right?
The "ERROR 404" is filler text for the embed.

src/config/get-config.js
naomi-lgbt: Okay, I can adjust this. 🙂

src/commands/suspend.js
naomi-lgbt: I could have it catch an error, but the idea was that the command should still suspend the target even if the channel creation setting isn't configured correctly.

src/commands/suspend.js
naomi-lgbt: Yeh I wrote the code such that as long as the bot can find the "suspended" role and the target, it will do so - everything else will generate an error if not configured correctly but still perform the role assignment.

src/commands/suspend.js
naomi-lgbt: Realistically, the `user.roles.add` shouldn't fail at all - `user` and `suspend` have both been checked. 
We don't want the `user.roles.remove` to fail the process as this would fail every time we suspended someone who didn't have a role. Removing the roles is part of the process in case the user has a role that would override the suspended role's settings.

src/commands/suspend.js
naomi-lgbt: I wrote this check as a client message because it indicates that the bot is still alive, but I can turn it to  a silent console.error if you'd rather.

src/commands/suspend.js
naomi-lgbt: I'll reword this to remove the "ERROR" language, though.

src/config/get-config.js
naomi-lgbt: This is done.

src/commands/suspend.js
naomi-lgbt: Changed it to error out if the category is missing.

src/commands/suspend.js
naomi-lgbt: Changed it to error out if the channel is missing.

src/commands/suspend.js
naomi-lgbt: Done

src/commands/suspend.js
naomi-lgbt: Done

src/commands/suspend.js
naomi-lgbt: Changed to a `console.log`

naomi-lgbt:
Okay, I've moved all of the environment variable checks to the top of the function, and the command will `console.log` a message and terminate if any of the variables are not set properly. 

I also added the optional DM feature and a close command.

src/commands/close.js
```diff
@@ -0,0 +1,23 @@
...
+    //check for user permissions
+    if (!message.member.hasPermission('MANAGE_CHANNELS')) {
+      console.log('Missing permissions.');
+    }
+    if (!log) {
+      console.log('Log channel not found.');
+      return;
+    }
+    target.delete().catch((e) => console.error(e));
+    log.send(`${message.author} deleted a channel.`);
```
bradtaniguchi: Do we want this to fail silently? If we are adding `.catch` to things that can fail we can always add a try/catch around the entire command function instead, so we still log the error, but don't send a message back saying the channel was deleted when it wasn't.

src/commands/suspend.js
```diff
@@ -0,0 +1,96 @@
...
+        {
+          name: 'Reason',
+          value: `${reason}`
+        }
+      )
+      .setFooter('Please remember to follow our rules!');
+    modChannel.send(restrictEmbed);
+    //assign roles
+    user.roles.remove(user.roles).catch((e) => console.error(e));
+    user.roles.add(suspend).catch((e) => console.error(e));
```
bradtaniguchi: Same as the other comment, this can still fail silently in that we will say that we removed/updated the user(s) but they could fail, which is mis-leading. We could say nothing and log the actual problem, or return a message.

bradtaniguchi (commented):
Added 2 comments about the `.catch`s still potentially misleading us into things are happening when they aren't, where an action could fail, be caught and logged, but the "success" message will be sent in chat. 

I haven't tested these changes on my side yet (will to it later today or tomorrow if I have time) tho

src/commands/suspend.js
naomi-lgbt: Realistically, the only reason either this or the above comment should throw any error is if the bot itself doesn't have the correct permissions. The command code checks for everything else and terminates if a configuration is missing or a condition isn't met. I could look at adding a bot permission check if we think that's necessary - otherwise, yeah we could wrap the entire function in a try block and catch the errors all at once.

src/commands/close.js
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      const target = message.channel;
+      //check for log channel
+      const log = message.guild.channels.cache.find(
+        (channel) => channel.name === config.LOG_MSG_CHANNEL
+      );
+      //check for user permissions
+      if (!message.member.hasPermission('MANAGE_CHANNELS')) {
+        console.log('Missing permissions.');
+      }
```
bradtaniguchi: ### This doesn't seem to do anything if the user doesn't have permissions to manage channels allowing anyone to close any channel by calling this command.

src/commands/close.js
```diff
@@ -0,0 +1,27 @@
...
+      if (!log) {
+        console.log('Log channel not found.');
+        return;
+      }
+      target.delete();
```
bradtaniguchi: This command would close **any** channel right? Isn't that too extreme? Shouldn't we have some level of checks somewhere to prevent us from randomly closing any channel? 

If we open "temp" channels for suspended users, we should be able to recognize that channel is for suspended users via:
- the name of the channel should have "suspended"
- the channel should be under the "suspended" room, from the config.

bradtaniguchi (changes requested):
Pointed out some security issues related to the close function.

src/commands/close.js
naomi-lgbt: Oops! Forgot the return statement 😅

naomi-lgbt: This is corrected

src/commands/close.js
naomi-lgbt: Command now checks if the channel name includes "suspended", and if the channel category name includes the SUSPEND_CATEGORY value from config. 
If either of these fails, it logs it and ends the command.

src/commands/suspend.js
```diff
@@ -0,0 +1,98 @@
+const Discord = require('discord.js');
+const getConfig = require('../config/get-config.js');
+const config = getConfig();
+module.exports = {
+  prefix: 'suspend',
+  description: 'Suspends a user for the given reason',
```
bradtaniguchi: During testing I think we should update this description to include something mentioning only admins have access, to this command. 

In the future we can not even show this action if a user isn't an admin, but that can be done later.

also providing an example might be useful

src/commands/suspend.js
```diff
@@ -0,0 +1,98 @@
...
+          },
+          {
+            name: 'Reason',
+            value: `${reason}`
+          }
+        )
+        .setFooter('Please remember to follow our rules!');
+      modChannel.send(restrictEmbed);
+      //assign roles
+      user.roles.set([suspend]);
```
bradtaniguchi: Add an `await` here
1. To unwrap any HTTP errors we get on assignment due to lack of permissions
2. allow us to "wait" for the assignment as the rest of the code doesn't work if we can't assign the role

bradtaniguchi (commented):
We need to update the README.md with the updated permissions, as after testing it on my side I got errors (after the async/await issue I mentioned was fixed) related to the bot missing permissions.

New permissions the bot needs (along with the existing one)
- manage roles
- manage channels

src/commands/suspend.js
```diff
@@ -0,0 +1,98 @@
...
+      //create suspend channel
+      const channelName = `suspended-${user.user.username}`;
+      message.guild.channels.create(channelName, {
```
bradtaniguchi: Adding `await` here also fixes uncaught errors if we didn't add permissions for channel creation

src/commands/close.js
```diff
@@ -0,0 +1,35 @@
+const getConfig = require('../config/get-config.js');
+const config = getConfig();
+module.exports = {
+  prefix: 'close',
+  description: 'Closes the channel.',
```
bradtaniguchi: Update this to be more descriptive about its use, and add a mention only admins can use this on "suspended" channels.

Like the other command this will change according to who calls this command, but this can be done later.

bradtaniguchi (commented):
I think when the "suspended" channel is created, the bot should add a message in the room specifying that `FCC close` can be used to close the suspended room. 

Also does the bot get added to the generated suspended channels?

src/commands/suspend.js
naomi-lgbt: Description has been updated to clarify that only admins have access, and to provide syntax.

src/commands/suspend.js
naomi-lgbt: Added await.

src/commands/suspend.js
naomi-lgbt: Added await.

src/commands/close.js
naomi-lgbt: Added clarification: only admins can access, can only target temporary "suspended" channels.

naomi-lgbt:
> I think when the "suspended" channel is created, the bot should add a message in the room specifying that `FCC close` can be used to close the suspended room.
> 
> Also does the bot get added to the generated suspended channels?

The bot now sends the requested message to the channel.

The bot does not get automatically added to the generated channel - I played around with getting that to work, but could not. Tried exporting the `client` from the index.js file, with no luck. However, the bot **should** have access to the channel automatically as the `@everyone` role gets restricted but Discord applies a separate role to the bot that should supercede that restriction. So unless that role has been manually changed, there should be no issue.

naomi-lgbt:
> We need to update the README.md with the updated permissions, as after testing it on my side I got errors (after the async/await issue I mentioned was fixed) related to the bot missing permissions.
> 
> New permissions the bot needs (along with the existing one)
> 
> * manage roles
> * manage channels

I can redo the link example and add these permissions to that block. In connection to the comment above, though, I think we also need to add the "read message history", "read/see text voice channels", and "send messages" permissions so the bot will automatically join the suspended channels. The README includes instructions for how to restrict the bot from a specific channel, and including these permissions won't affect that process.

src/commands/close.js
```diff
@@ -0,0 +1,36 @@
+const getConfig = require('../config/get-config.js');
+const config = getConfig();
+module.exports = {
+  prefix: 'close',
+  description:
+    'Closes the channel. This command requires admin privliges, and will only work on the automatically created "suspended" channels.',
```
bradtaniguchi: Typo with the word "privliges", should be "privileges"

If your using VSCode, I use an extension called [code spell checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker), its a life saver as I can't spell worth a damn.

bradtaniguchi (commented):
> The bot does not get automatically added to the generated channel - I played around with getting that to work, but could not. Tried exporting the client from the index.js file, with no luck. However, the bot should have access to the channel automatically as the @everyone role gets restricted but Discord applies a separate role to the bot that should supercede that restriction. So unless that role has been manually changed, there should be no issue

So right now there isn't a way to have the bot add itself to the room? Or we aren't sure and it just isn't working (I wasn't able to get it to work on my server either)
After I manually added the bot to the suspended room and ran `FCC close` it worked as expected for an admin, and the non-admin user.

I noticed a few things.
1. After closing the room the user doesn't become "un-suspended" they are still in that role, should we unsuspend them? or let the admin handle it manually? Might want to ask what should the behavior be from others.

2. The log message is rather plain, compared to the "suspended" message with just:
```
@bradtaniguchi deleted a channel.
```
Maybe we can add an embedded message that says something similar? Might also want to ask for feedback on what it should say.

src/commands/close.js
naomi-lgbt: Oof. Big embarrass lol. Will fix that real quick.

naomi-lgbt:
I could try doing an embed for the close channel message - I couldn't get the bot to send the channel name though. Discord kept changing it to "deleted channel" because the channel no longer exists. That might not happen with an embed though. 

I thought about removing the role automatically, but thought it should remain manual - if the discussion goes poorly and we choose to ban/leave them suspended, closing the channel shouldn't interfere with that option. I **could** do an "unsuspend" command separately, but that might be a little much.

naomi-lgbt:
Okay, I've modified the close command. It now sends an embed with the deleted channel's name and a decision status that the appeal was not approved. Additionally, it takes an optional parameter of a user tag - if the tag is present, the bot removes the suspended role from the user and changes the embed to state the appeal WAS approved.

bradtaniguchi:
Further testing on my side "from scratch" made me realize something that is actually really annoying if your not careful:
image
The **middle** permission specifically says:
"Members with this permission can create new roles and edit/delete roles lower than this one", 

This means the **order of the roles in the discord server needs to have the bot role higher than the suspended role**. We should add this to the README somewhere as its a gotcha, and I spent easily half an hour trying to figure the issue out XD. So this means the order of the roles in the "roles" panel needs to be the "most power" at the top, and `@everyone` on the bottom, where suspended should also be lower than everyone (to be more "correct" than anything).

src/commands/suspend.js
```diff
@@ -0,0 +1,102 @@
...
+          {
+            id: message.guild.id,
+            deny: ['VIEW_CHANNEL', 'READ_MESSAGE_HISTORY', 'SEND_MESSAGES']
+          }
+        ],
+        parent: category
+      });
+      await suspendChannel.send(
+        `This channel has been created for ${user} to discuss their suspension from the server. Once the discussion has concluded, an admin may use the \`close\` command to automatically close this channel.`
+      );
```
bradtaniguchi: This errors out due to the bot not having permissions in this channel. Comment it out for now until we fix the other issue where we can't get the bot into the suspendChannel.

I think you already mentioned that we need the `client` passed to this function so we can go `client.user.id` (which is the bot I believe). We can add this during the TS refactor, so then we can un-comment this code and should be good to go.

README.md
```diff
@@ -76,9 +76,11 @@ Just beside your bot's icon, you will see your `CLIENT ID`. Copy the `CLIENT ID`
 
 Then replace the `YOUR_CLIENT_ID_HERE` with your bot's `CLIENT ID` in the link provided below. After that, you can use the link to add the bot to any existing or newly created Discord server (as long as you have the `Manage Server` permission).
 
-_https://discordapp.com/oauth2/authorize?client_id=YOUR_CLIENT_ID_HERE&scope=bot&permissions=8192_
+_https://discordapp.com/oauth2/authorize?client_id=YOUR_CLIENT_ID_HERE&scope=bot&permissions=268512272_
 
-**IMPORTANT NOTE :** When you click on the link and want to add the bot to a server, it'll ask for the `Manage Messages` permission. You have to grant the bot this permission or it will not work properly.
+**IMPORTANT NOTE :** When you click on the link and want to add the bot to a server, it'll ask for the following permissions. You must approve these permissions for the bot to work correctly:
+
+`Manage Roles`, `Manage Messages`, `Manage Channels`, `View Channels`, `Read Message History`, `Send Messages`
```
bradtaniguchi, line 84: Add another important note:
```
**IMPORTANT NOTE :** When setting up the roles within the server including the `suspended` role for the suspended feature for admins, verify the "bot role" that is **above** the suspended role, otherwise the bot will be unable to manage the `suspended` role.
```

src/commands/suspend.js
naomi-lgbt: Commented this out for now.

README.md
naomi-lgbt: Have reworded this. :)

bradtaniguchi (approved):
LGTM :+1: :shipit: :rocket: 

**note** there still might be some errors related to the bot not being "moved" into the room regardless of permissions (I could never get this to work, but @nhcarrigan saw the designed behavior). This can be done after the TS refactor

PR #72, opened 2020-06-24
Added editorconfig as a Dependency

JoshQuaintance:
I noticed that there is a `.editorconfig` on the repo, but it won't work without the depedency. Just adding it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "dependencies": {
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
-    "editorconfig": "^0.15.3",
     "express": "^4.17.1",
     "highlight.js": "^10.0.3",
     "prettier": "^2.0.5",
@@ -26,7 +25,8 @@
     "jest": "^26.0.1",
     "lint-staged": "^10.2.2",
     "nodemon": "^2.0.3",
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "editorconfig": "^0.15.3"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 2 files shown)


bradtaniguchi:
I realized this "works" on my end on my machines when I use VSCode with [this extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig). I'm not 100% sure what the npm package does, but I don't think we need it if you use any of the following editors listed as this site:

https://editorconfig.org/#download

naomi-lgbt:
And if you have the necessary plugin. I don't, apparently - which explains why I was battling that line break rule for so long. 
Adding the editorconfig dependency seems to do the same thing as adding the extension, so including it in the package.json would just help catch contributors who don't have the plugin.

bradtaniguchi:
So with the npm extension the editorconfig magically runs with whatever editor your using? (or at least the non-plugin equiped vscode)?

package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
+    "editorconfig": "^0.15.3",
```
bradtaniguchi: Move this to `devDependencies` since it isn't directly used by the bot

naomi-lgbt:
Hmm, maybe not. I tested it out more thoroughly and it does not seem to automatically change my linebreak to LF.

package.json
JoshQuaintance: Ok

JoshQuaintance:
> Hmm, maybe not. I tested it out more thoroughly and it does not seem to automatically change my linebreak to LF.

You need to Save. It changes stuff on save.

bradtaniguchi (approved):
If this works on save, and works for those on other systems (I'm on Linux), then should be good!

PR #77, opened 2020-06-26
Add gitpod support

Twaha-Rahman:
This PR adds support for GitPod in the repo(with MongoDB).

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,5 +1,8 @@
+image:
+  file: .gitpod.Dockerfile
 # Command to start on workspace startup
 tasks:
+  - command: mkdir -p /workspace/data && mongod --dbpath /workspace/data
   - init: npm i
     command: npx nodemon
 
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -2,9 +2,8 @@ image:
   file: .gitpod.Dockerfile
 # Command to start on workspace startup
 tasks:
-  - command: mkdir -p /workspace/data && mongod --dbpath /workspace/data
-  - init: npm i
-    command: npx nodemon
+  - init: mkdir -p /workspace/data && mongod --dbpath /workspace/data
+  - init: npm ci
 
 vscode:
   extensions:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,9 @@
+# Command to start on workspace startup
+tasks:
+  - init: npm i
+    command: npx nodemon
+
+vscode:
+  extensions:
+    - esbenp.prettier-vscode@5.1.0:M3Ww4EayZWhGCgk49C1Ldw==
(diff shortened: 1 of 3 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,12 @@
+image:
+  file: .gitpod.Dockerfile
+# Command to start on workspace startup
+tasks:
+  - command: mkdir -p /workspace/data && mongod --dbpath /workspace/data
+  - init: npm i
+    command: npx nodemon
+
```
bradtaniguchi: When testing out this PR I got errors in my console due to `npx nodemon` running the server and failing. We shouldn't have to have any npm script. 

I can't tell if the mongodb command works as expected if its ran every time the workspace is loaded, but shouldn't
 that also be an `init` rather than a command?

Regardless the `npm i` can be changed to `npm ci` to prevent random re-writes to `npm install`, which could throw some devs off. 

So I suggest the following:
```yaml
tasks:
  - init: mkdir -p /workspace/data && mongod --dbpath /workspace/data
  - init: npm i
```
(if gitpod doesn't allow multiple `init`s, then having adding another `&&` could be a workaround)

README.md
```diff
@@ -6,7 +6,11 @@ A Discord test bot used to test out the architecture and implementation of a Dis
 
 ## Contributing
 
-So, you want to contribute to this repo? We are glad to recieve PRs from anybody who is interested. Follow these steps if you are new to contributing to open source projects:
+So, you want to contribute to this repo? We are glad to recieve PRs from anybody who is interested.
+
+The easiest way to contribute to the development of the bot is using [Gitpod](http://gitpod.io/#https://github.com/bradtaniguchi/discord-bot-test). It starts up an already working version of the bot with all the code so you can start working right now! (You'll have to add the `TOKEN` variable in the `.env` file)
+
+Follow these steps if you are new to contributing to open source projects:
```
bradtaniguchi: I'm not sure if we should say "you can start working right now!" as the user still needs to setup the discord bot in discord. Maybe we should primarily mention how gitpod can be used as a development environment instead of running things locally. Contributors will still need to setup their own discord bot on their own end though to help contribute (and deal with all the annoying aspects of that side of things!)

.gitpod.yml
Twaha-Rahman: I think Gitpod allows multiple `init`

README.md
Twaha-Rahman: Hmmm... this seems reasonable. I think I'll throw in the Gitpod logo too.

bradtaniguchi: Should we wait for the gitpod logo, or should we be good to go without it? :)

Twaha-Rahman: Ooops! Completely forgot about this PR.... my apologies. We can add a logo quite easily... but do you think the README instructions are good?

bradtaniguchi: The current words should be good

Not a big deal BTW :D

bradtaniguchi (approved):
LGTM :+1:

PR #79, opened 2020-06-29
Fix/suspend command

naomi-lgbt:
While testing the TS conversion, I found everything to work as intended - but I ran in to the same permissions error with the bot not being able to access the automatically created `suspended-user` channel. Playing with permissions, the only Discord-side workaround is to give the bot Admin access.

Since that is not the route we want to go, I've found an alternative fix. I have updated the config and suspend command to look for a manually created bot role (set in the `.env` file) - this role is *different* from the role Discord automatically creates when you add a bot to a server. 

By setting this role, the bot gets permission to see the channel, and can then perform the close command as intended.


bradtaniguchi (approved):
This looks to have fixed the issue where the bot doesn't enter the created room! Nice find!

PR #80, opened 2020-06-30
Verbose config was incorrect

naomi-lgbt:
The `VERBOSE` config setting was looking for `PREFIX` to be true. Changed it to look for `VERBOSE` to be true.


bradtaniguchi (approved):
good catch!, LGTM :+1:

PR #81, opened 2020-06-30
Add forum command

naomi-lgbt:
Add a forum command that manually gets the latest 5 posts. 

I was playing with the idea of making this automatic, but was running into hiccup after hiccup. Could be something we add later, if we choose to go this route at all.

### DIFF
--- a/src/commands/forum.ts
+++ b/src/commands/forum.ts
@@ -6,34 +6,22 @@ export const forum: CommandDef = {
   prefix: 'forum',
   description: 'Gets the recent forum activity.',
   command: async (message: Message): Promise<void> => {
-    const data = await fetch('https://forum.freecodecamp.org/latest.json');
-    const parsed = await data.json();
-
-    const forumEmbed: MessageEmbed = new MessageEmbed()
-      .setTitle('Latest Forum Activity')
-      .setDescription('Here are the five most recent posts.')
-      .addFields(
-        {
-          name: `${parsed.topic_list.topics[0].title}`,
-          value: `[${parsed.topic_list.topics[0].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[0].id})`
-        },
-        {
-          name: `${parsed.topic_list.topics[1].title}`,
-          value: `[${parsed.topic_list.topics[1].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[1].id})`
-        },
-        {
-          name: `${parsed.topic_list.topics[2].title}`,
-          value: `[${parsed.topic_list.topics[2].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[2].id})`
-        },
-        {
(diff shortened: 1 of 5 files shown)


src/commands/forum.ts
```diff
@@ -0,0 +1,39 @@
...
+          value: `[${parsed.topic_list.topics[2].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[2].id})`
+        },
+        {
+          name: `${parsed.topic_list.topics[3].title}`,
+          value: `[${parsed.topic_list.topics[3].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[3].id})`
+        },
+        {
+          name: `${parsed.topic_list.topics[4].title}`,
+          value: `[${parsed.topic_list.topics[4].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[4].id})`
+        }
```
bradtaniguchi: This can be done in a loop, where we get only the first 5 items using something like `slice` or `splice` (I usually go with `slice` as it doesn't affect the starting array)

src/commands/forum.ts
```diff
@@ -0,0 +1,39 @@
...
+      );
+    message.channel.send(forumEmbed).catch((e) => console.error(e));
```
bradtaniguchi: Instead of having a `.catch` here, add a try/catch around the entire command callback incase there is an error with sending this message (doubt it) or with node-fetch

src/commands/forum.ts
naomi-lgbt: Changed to a slice and a forEach.

src/commands/forum.ts
naomi-lgbt: Done.

src/commands/forum.ts
```diff
@@ -6,34 +6,22 @@ export const forum: CommandDef = {
...
-        {
-          name: `${parsed.topic_list.topics[3].title}`,
-          value: `[${parsed.topic_list.topics[3].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[3].id})`
-        },
-        {
-          name: `${parsed.topic_list.topics[4].title}`,
-          value: `[${parsed.topic_list.topics[4].last_poster_username} replied on ${parsed.topic_list.topics[0].last_posted_at}](https://forum.freecodecamp.org/t/${parsed.topic_list.topics[4].id})`
-        }
+    try {
+      const data = await fetch('https://forum.freecodecamp.org/latest.json');
```
bradtaniguchi, line 11: Is it possible to add a TS interface (add it in another file somewhere you think would be useful for TS defs for api requests) for this data?

naomi-lgbt: Okay, I think I've done this correctly.

bradtaniguchi (approved):
code LGTM :+1: 

Didn't get to test it yet, but I think it should be fine (if not we can fix it haha)

I'm not sure of the other approaches you took, but one I see working without much trouble would be to make this api call, and get the list of these forum posts ever few minutes/hours. Not sure of what other api /files are available but we can build off of this command in the future :)

Issue #59, opened 2020-06-18
Update stats command to include more information about the bot. 


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
When running the `<prefix> stats` command, we should have the reply include more of the following:
- Uptime (time since the bot was started)
- Commit, tag and time the bot was deployed with/on
  - this should be available either from heroku/environment, or to be environment agnostic, available from a JSON file created during build and deployed with the bot itself that is loaded upon startup.

**Describe the solution you'd like**
Update the command with those features, might require updates to travis-ci to do this automatically before/after deployment

JoshQuaintance:
I can implement the Uptime and maybe the commit and tag. I can totally do the time the BOT is deployed at. This should be an easy task to add.

*Note: Do you want these to be a ts file?*

bradtaniguchi:
The `ts` changes aren't merged yet, but this feature shouldn't be reliant on it.

I'd create a script (in js is fine, it can be migrated later) that can be ran via an npm script that gets this information and creates a `json` file at the path of our choice. I'm honestly not sure if this script will be ran in heroku or in travis yet, your welcome to look into when we'd run this beyond just getting the actual git hash

The up-time feature should be vastly easier. If you want to just do that you can open a PR just for that one part, and we can do the commit/tag/time later as it depends on how the deployment/runtime environment.

naomi-lgbt:
As an FYI, I discovered yesterday that Heroku closes and reloads the bot every day, so the uptime will be reset.

JoshQuaintance:
Alright, I'll work on the uptime feature, as of the comit/tag/time, I might need to do some more research on how heroku or travis takes on data. Just for clarification, the uptime is going to rely on discord.js's Client.uptime right?

bradtaniguchi:
> As an FYI, I discovered yesterday that Heroku closes and reloads the bot every day, so the uptime will be reset.

That should be fine, the bot technically will "go down" every half hour or so of not being used on the free tier, so the uptime will be very short with our current setup

bradtaniguchi:
> Alright, I'll work on the uptime feature, as of the comit/tag/time, I might need to do some more research on how heroku or travis takes on data. Just for clarification, the uptime is going to rely on discord.js's Client.uptime right?

Yea that works, I didn't know about `Client.uptime` and would of done it totally manually, but either should work for the feature. Good find! :D

JoshQuaintance:
[...]

PR #82, opened 2020-07-01
Getting Commit Hash Locally and When Deployed on Heroku Closes (#59)

JoshQuaintance:
If `heroku buildpacks:add https://github.com/ianpurvis/heroku-buildpack-version -a [app-name]` was successfully added by nhcarrigan to the bot app, the variable should be available to carry when deployed. If the user is using the bot locally, then it will get the commit hash locally.

### DIFF
--- a/src/commands/stats.ts
+++ b/src/commands/stats.ts
@@ -1,5 +1,9 @@
 import { getUpTime } from '../utilities/get-up-time';
 import { CommandDef } from './command-def';
+import getRepoInfo from 'git-repo-info';
+const info = getRepoInfo();
+import dotenv from 'dotenv';
+dotenv.config();
 
 export const stats: CommandDef = {
   prefix: 'stats',
@@ -12,6 +16,13 @@ export const stats: CommandDef = {
   command: (message, client): void => {
     try {
       const uptime = getUpTime(client);
+
+      // If run locally, it will get the commit hash using a different approach
+      const versionHash = process.env.SOURCE_VERSION;
+      const commitHash =
+        versionHash == undefined
+          ? info.abbreviatedSha
+          : versionHash.slice(0, 10);
 
       const statsEmbed = {
         color: '#0099FF',
@@ -29,6 +40,10 @@ export const stats: CommandDef = {
           {
             name: 'Bot Online Time',
             value: 'Wake Time from JSON file'
+          },
+          {
+            name: 'Version',
+            value: `[${commitHash}](https://github.com/bradtaniguchi/discord-bot-test/commit/${commitHash})`
           },
           {
             name: 'Created on',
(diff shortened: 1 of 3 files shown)


src/commands/stats.ts
```diff
@@ -1,5 +1,9 @@
 import { getUpTime } from '../utilities/get-up-time';
 import { CommandDef } from './command-def';
+import getRepoInfo from 'git-repo-info';
+const info = getRepoInfo();
+import dotenv from 'dotenv';
+dotenv.config();
```
bradtaniguchi: `dotenv` should be ran earlier in the `index.ts` or `get-config` file. I think during the refactor, or earlier this wasn't being ran as early as possible. Instead of "hiding" this call away in this command, we should move it somewhere in the `index.ts` file (earlier in the imports, the "earlier" it runs)

src/commands/stats.ts
```diff
@@ -13,6 +17,13 @@ export const stats: CommandDef = {
     try {
       const uptime = getUpTime(client);
 
+      // If run locally, it will get the commit hash using a different approach
+      const versionHash = process.env.SOURCE_VERSION;
+      const commitHash =
+        versionHash == undefined
+          ? info.abbreviatedSha
+          : versionHash.slice(0, 10);
```
bradtaniguchi: I guess the `SOURCE_VERSION` will be found when running locally, otherwise the `info.abrviatedSha` will be loaded? Isn't `info` always available? Or does this not work in some environments/situations?

this could probably could be done with just:
```ts
const commitHash = process.env.SOURCE_VERSION || info.abbreviatedSha;
```

src/commands/stats.ts
```diff
@@ -30,6 +41,10 @@ export const stats: CommandDef = {
             name: 'Bot Online Time',
             value: 'Wake Time from JSON file'
           },
+          {
+            name: 'Version',
+            value: `[${commitHash}](https://github.com/bradtaniguchi/discord-bot-test/commit/${commitHash})`
```
bradtaniguchi, line 41: I was just thinking about having the sha shown as is, but having a link is better. Good idea!

src/commands/stats.ts
JoshQuaintance: `SOURCE_VERSION` will be found when it's deployed in heroku. But, I will fix this. I just never noticed that if the value if `undefined` or `null` it will go to the fallback.

src/commands/stats.ts
JoshQuaintance: Actually, we can just run the `process.env.SOURCE_VERSION` inside the `get-config` file. Nice, I totally forgot about this

JoshQuaintance: Actually, it will work fine without importing it again. So deleting these 2 lines will be fine since it was called before.

naomi-lgbt:
Closes #59

bradtaniguchi (approved):
LGTM :+1:

just tested it out locally and worked fine! Will have to double check once deployed :)

PR #83, opened 2020-07-01
Feature/54/update thanks command

### LINKED ISSUE #54: Dramatically update thanks parsing capabilities
Apparently the original gitter bot's "thanks" command supported multiple languages and even typos.

We can carry over a similar solution from the original. We should also add the same tests to verify we get the same support as before.

**Additional context**
Here is the regex that was applied, it was slightly hard to find, but I guess this is what was used to check if the message had the command:
https://github.com/freeCodeCamp/camper-gitter-bot/blob/529cc6e989f2f9661e3dd65d927727c6489c8249/data/rooms/RoomMessages.js#L49

naomi-lgbt:
Changed the thanks command to use the regex from the gitter bot. 

Closes #54


src/tests/thanks.test.ts
```diff
@@ -24,13 +24,13 @@ describe('thanks', () => {
           mentions: { users: new Map().set('86890631690977280', {}) }
         } as Partial<Message>) as never)
       ).toEqual(true));
-    test('returns true if thanking self with THANK YOU', () =>
+    test('returns false if thanking self with THANK YOU', () =>
```
bradtaniguchi: Ah I remembered why this is the case. There is another check in the file if the user is checking themselves, which will return a different message ("can't thank self").

Technically the regex change doesn't even know if the user is "thanking themselves", so this missing test-case is actually because the regex wont recognize `THANK YOU <name of user>` self or not.

src/commands/thanks.ts
```diff
@@ -1,5 +1,6 @@
 import { Message } from 'discord.js';
 
+const thankregex = /((?:^|\s)(?:(?:th(?:n[qx]|x)|t[xyq]|tn(?:[x]){0,2})|\w*\s*[.,]*\s*than[kx](?:[sxz]){0,2}|than[kx](?:[sxz]){0,2}(?:[uq]|y(?:ou)?)?)|grazie|arigato(?:[u]{0,1})|doumo|gracias?|spasibo|dhanyavaad(?:hamulu)?|o?brigad(?:o|a)|dziekuje|(?:re)?merci|multumesc|shukra?an|danke)\b/gi;
```
bradtaniguchi: Rename variable to `thankRegex`

src/tests/thanks.test.ts
naomi-lgbt: image

This may have been the case with gitter, but discord.js has a check to see if the mentioned user is the same as the message author.

src/commands/thanks.ts
naomi-lgbt: Changed.

src/tests/thanks.test.ts
bradtaniguchi: That check is somewhere else in our code:
https://github.com/bradtaniguchi/discord-bot-test/blob/40ab1a7dc56e05134dccb5b4c31660348430dfe8/src/commands/thanks.ts#L15

naomi-lgbt: Ah, I see.
So this test should return true just because the regex passes and the message contains a mention. 
Any idea why it doesn't return true?
I tested the regex and it does match...

src/commands/thanks.ts
```diff
@@ -1,5 +1,6 @@
 import { Message } from 'discord.js';
 
+const thankRegex = /((?:^|\s)(?:(?:th(?:n[qx]|x)|t[xyq]|tn(?:[x]){0,2})|\w*\s*[.,]*\s*than[kx](?:[sxz]){0,2}|than[kx](?:[sxz]){0,2}(?:[uq]|y(?:ou)?)?)|grazie|arigato(?:[u]{0,1})|doumo|gracias?|spasibo|dhanyavaad(?:hamulu)?|o?brigad(?:o|a)|dziekuje|(?:re)?merci|multumesc|shukra?an|danke)\b/gi;
```
bradtaniguchi: Move to be declared in function :)

naomi-lgbt: Done :)

src/tests/thanks.test.ts
naomi-lgbt: This is also fixed.

bradtaniguchi (approved):
LGTM :+1:

PR #84, opened 2020-07-01
Fix/husky lint ts

naomi-lgbt:
The husky pre-commit lint is not configured to read the TS files, currently. This change ensures that TS files get linted when committing.


bradtaniguchi (approved):
Good catch!

PR #86, opened 2020-07-01
Feature/65/update formatting

### LINKED ISSUE #65: Refactor code formatting feature to remove "feedback" step
Currently when adding the :robot: emoji on a message the bot tries to format the given code block using hightlightjs and prettier. If highlighjs gives 2 guesses back, the bot asks for clarification with 2 suggestions. This is slightly annoying and makes the feature harder to use, when most of the time the first guess is "right" anyways.

Remove the feedback step so the bot just uses the first guess it has, and continues to format/setup the correct highlighted language just as before.

**Additional context**
This should be done **after** the migration to TS, see #55

### LINKED ISSUE #66: Add "format" command
Due to changes with the "reaction" code format feature, the bot will be more aggressive on auto-guessing the language. To provide a more "end-user" version that will provide more support for different languages and potential prettier support

Add a new command `<prefix> format <message_link> <language>` command that will take a given message contents, the code in that message and try to format+apply syntax highlighting to that message.

This can use some of the existing utilities for the :robot: code format feature.

**Additional context**
Prettier should only be applied if the language given is actually recognized by prettier, otherwise don't try to format the code.

### LINKED ISSUE #67: Add specific case for HTML+JS language format guesses to format for XML
a code block like the following:
```xml
<script>
let me = "Awesome"
let you = "Better"
</script>
<div id="green>
<p id="blue>
</p>
</div>
```
can be formatted using XML. This can be used **if** the language guesses provided by highlightjs are HTML and JS (in that order). 

See above, this would make JS in HTML syntax highlighting better for :robot: formatted code.

naomi-lgbt:
- [X] Removed the confirmation message - command now runs with the first guess automatically. 
- [X] Configured the command to use XML if it detects HTML and either CSS or JS.
- [X] Created a `format` command that takes the message ID as the parameter, and returns a formatted message. 

Closes #65, closes #66, and closes #67

@Twaha-Rahman - You wrote this feature originally, I believe. Wouldn't mind your eyes on this change too. 🙂


src/commands/format.ts
```diff
@@ -0,0 +1,21 @@
...
+    if (!messageID || isNaN(parseInt(messageID))) {
+      message.channel.send('Invalid syntax');
+      return;
+    }
+    const messageContent = await message.channel.messages.fetch(messageID);
+    if (!messageContent) {
+      message.channel.send('Message not found');
+      return;
+    }
+    return addFormatting(messageContent);
```
bradtaniguchi: Add try/catch block around this to log/capture any weird errors.

src/commands/format.ts
```diff
@@ -0,0 +1,21 @@
+import { CommandDef } from './command-def';
+import { Message } from 'discord.js';
+import { addFormatting } from './add-formatting';
+
+export const format: CommandDef = {
+  prefix: 'format',
+  description: 'Formats the code contained in the message ID.',
+  command: async (message: Message): Promise<void> => {
+    const messageID = message.content.split(' ')[2];
```
bradtaniguchi: Verify the amount is as expected, as if the content passed isn't long enough this code will explode.

src/commands/add-formatting.ts
```diff
@@ -49,28 +49,17 @@ export async function addFormatting(message: Message): Promise<void> {
...
+    (languageGuesses.includes('CSS') || languageGuesses.includes('JavaScript'))
+  ) {
+    languageGuesses[0] = 'XML';
+  }
+  const supportedLanguage = isSupportedByPrettier(languageGuesses[0]);
+  if (supportedLanguage) {
+    const formattedCode = formatter(content, supportedLanguage);
+    message.channel.send(formatCodeBlock(supportedLanguage, formattedCode));
+  } else {
+    message.channel.send(formatCodeBlock(languageGuesses[0], content));
```
bradtaniguchi: Use early return in the first `if` statement rather than an `else` statement. (Sorry I don't like else ;D)

src/commands/format.ts
naomi-lgbt: Done

src/commands/format.ts
naomi-lgbt: Added a check for length

src/commands/add-formatting.ts
naomi-lgbt: I like else :P
Done! :)

src/commands/format.ts
```diff
@@ -0,0 +1,29 @@
+import { CommandDef } from './command-def';
+import { Message } from 'discord.js';
+import { addFormatting } from './add-formatting';
+
+export const format: CommandDef = {
+  prefix: 'format',
+  description: 'Formats the code contained in the message ID.',
```
bradtaniguchi: Where and how is the user suppose to get the messageId? My original understanding was that this function would load via a discord URL?

naomi-lgbt: So...
That was the original plan.
But discord.js doesn't have a way to find a message by the url, that I could see.
So I went with this route instead.
But just now I've realised how I could parse the ID *from* the URL, and I'll go ahead and try to factor that into the command instead. Allow me a few minutes and I'll have this changed. 

Also, the ID can be copied just like the url (right-click drop down) if you've got dev mode turned on 🙂

naomi-lgbt: This is changed now.

bradtaniguchi (approved):
LGTM :+1: 
I'll wait for @Twaha-Rahman for a big if they want to review these changes aswell.

bradtaniguchi:
I think its fine to merge this for now without any extra review from Twaha-Rahman

PR #87, opened 2020-07-02
Uptime and Wake Time Closes (#59)

### LINKED ISSUE #59: Update stats command to include more information about the bot.

JoshQuaintance:
This will close #59 
- [x] Uptime will display the time the bot has been up for, `[DD] [HH] [MM] [SS]` format.
- [x] Bot Wake Up Time will show the time the bot was up at. `[HH:MM] [AM/PM] [TimeZone]` format. It is also linked to a google search for the time to show the relative time to the user's area. 

It turns out we didn't need the JSON file after all. Just a simple export from the start. The Wake Time will open up a google search to their relative location since the bot will only give the time from where it's located. So I just rigged up a google search query and let Google do the job for me 😆. 
Here's an example: 
image

**Note: Ignore most of the commits, most of them are me Trying to merge the upstream to my master**

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@types/luxon": "^1.24.1",
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
@@ -28,6 +27,7 @@
     "@types/express": "^4.17.6",
     "@types/highlightjs": "^9.12.0",
     "@types/jest": "^26.0.0",
+    "@types/luxon": "^1.24.1",
     "@types/node-fetch": "^2.5.7",
     "@types/prettier": "^2.0.1",
     "@types/strip-ansi": "^5.2.1",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "dependencies": {
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
-    "editorconfig": "^0.15.3",
     "express": "^4.17.1",
     "highlight.js": "^10.0.3",
     "prettier": "^2.0.5",
@@ -26,7 +25,8 @@
     "jest": "^26.0.1",
     "lint-staged": "^10.2.2",
     "nodemon": "^2.0.3",
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "editorconfig": "^0.15.3"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
+    "git-repo-info": "^2.1.1",
     "highlight.js": "^10.0.3",
     "node-fetch": "^2.6.0",
     "prettier": "^2.0.5",
--- a/package.json
+++ b/package.json
@@ -13,11 +13,13 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@types/luxon": "^1.24.1",
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "git-repo-info": "^2.1.1",
     "highlight.js": "^10.0.3",
+    "luxon": "^1.24.1",
     "node-fetch": "^2.6.0",
     "prettier": "^2.0.5",
     "strip-ansi": "^6.0.0"
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "discord.js": "^12.2.0",
     "dotenv": "^8.2.0",
+    "editorconfig": "^0.15.3",
     "express": "^4.17.1",
     "highlight.js": "^10.0.3",
     "prettier": "^2.0.5",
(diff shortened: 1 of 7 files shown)


bradtaniguchi (commented):
The google search link isn't necessary, since I don't see anyone wanting to get any more info for a given timestamp.

The `onlineAt` could be defined within the `get-config` function, which **should** be refactored to be passed to all the commands. (I'll try to do this later today)


After this I think we should update the date parsing to be cleaner, as right now the solution has gotten messier with multiple regex and custom date logic, when such problems have already been solved with existing date libraries. Such as luxon.

I recommend getting luxon setup, and migrating the `bot-online-time` function to it. After updating the command-def so more stuff is passed to the function, we can update this code to get the `onlineAt`/` Bot Online Time` information from the config itself.

JoshQuaintance:
I have removed the google search link since it's really unnecessary.

The `onlineAt` is moved into `get-config`.

The `getOnlineAt()` function is now utilizing the luxon package. Now it is cleaner using the object wrapper and formatting from luxon. 

The `stats.ts` file will now use the config to get the Online Time.

package.json
```diff
@@ -13,11 +13,13 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@types/luxon": "^1.24.1",
```
bradtaniguchi: Install as `devDepedency`

JoshQuaintance: Done

bradtaniguchi:
Wait for #88 To be review+merged, then we can use the `config` object that is initially set.

bradtaniguchi:
@JoshuaPelealu, with the updates in master, you should be able to get the `config` object in the function callback:
Here is an example from the `close` command:
https://github.com/bradtaniguchi/discord-bot-test/blob/master/src/commands/close.ts#L10

So you can slightly refactor the code to remove the extra `getConfig` call, and then we should be good to go :smile:

JoshQuaintance:
The function should now get 2 arguments inside the parameter and the import is not used.

src/commands/stats.ts
```diff
@@ -34,7 +34,7 @@ export const stats: CommandDef = {
           },
           {
             name: 'Bot Online Time',
-            value: 'Wake Time from JSON file'
+            value: config.ONLINE_AT
           },
           {
             name: 'Version',
```
naomi-lgbt, line 40: A quick question here - is hard-coding the repository URL the best route? Was testing these changes, and if I've done a commit on my repo, the bot pulls that commit hash but tries to link to this repo. Which works, because my repo is a fork? Just curious if the link should be a config item instead.

src/utilities/bot-online-time.ts
```diff
@@ -1,11 +1,8 @@
...
+
 export function getBotOnlineAt(): string {
-  const date = new Date();
-  const hour = date.getHours() < 10 ? `0${date.getHours()}` : date.getHours();
-  const mins =
-    date.getMinutes() < 10 ? `0${date.getMinutes()}` : date.getMinutes();
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const tz = /\((.*)\)/.exec(new Date().toString())![1];
-  const am_pm = hour >= 12 ? 'PM' : 'AM';
+  const dt = DateTime.local();
```
naomi-lgbt, line 4: Another question, while I'm testing this. 

Using the local format means the time on my instance is different than the times I saw when you were testing your instance. When this is deployed to Heroku, do we know that the local format will be compatible?

JoshQuaintance: Yes it will be compatible. So depending on where it's at, it will always change the time to the local area of where the bot is, but the format will always be in `[HH]:[MM]:[SS] [AM/PM] [TimeZone]`. 

Since it's now using luxon, it will use the object that is given by luxon and just make it into a string.

naomi-lgbt: Here's hoping the bot doesn't end up in the middle of the Atlantic Ocean then 🙃

src/commands/stats.ts
JoshQuaintance: There is a way to make it link it's own repository, but only locally. If it's going to be deployed on heroku, there is no way of finding where the repository is at. 
But, making it a config item would be a great idea. Ask the user which repository they want it linked to. 
Or even, we can grab the repository from the `package.json` as a fallback, then if they didn't specify any, it will just be Brad's repository that is linked.

JoshQuaintance: We can try asking Brad, if he just wants it hard-coded, or as a config file, or getting it from the `package.json` or all of them with fallbacks.

bradtaniguchi: The simplest solution to this would be to grab the information from the `package.json` on load of the script, however the problem wont exist in production environments, only in forked repos which is to be expected.

I think it should be **ok** to leave as is, to keep complexity down for the sake of simplicity. :)

naomi-lgbt: Sounds good to me!

naomi-lgbt (approved):
Ran these changes through a few tests and everything seems to work as intended.

bradtaniguchi (approved):
LGTM :+1:

PR #88, opened 2020-07-03
Update command-def to include config and client

bradtaniguchi:
- Update the command-def interface to include the single passed config. This should fix issues where getConfig is ran multiple times.

### DIFF
--- a/src/commands/bootstrap.ts
+++ b/src/commands/bootstrap.ts
@@ -157,7 +157,7 @@ export function bootstrap({
       }
       // Execute command
       try {
-        commands.get(commandArgument)?.command(message, client);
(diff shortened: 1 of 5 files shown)


bradtaniguchi:
@nhcarrigan Can you double check the changes in this PR, just in-case the changes broke something :)

naomi-lgbt:
Gave this a look - everything still seems to function as intended. Ran all the modified commands, all good there. Also checked by passing the wrong suspend channel into the config file, and the bot reacted accordingly.

naomi-lgbt:
There is one thing I want to test, still.

Need to see if updating the config works without a rebuild.

Edit: This works. 

LGTM

naomi-lgbt (approved):

PR #89, opened 2020-07-05
Add pin message feature

### LINKED ISSUE #78: Add "self pin" feature
Currently there is a way for admins/mods to pin messages to a channel using the pin message feature. However normal users that would like to "bookmark" or keep track of posts aren't able to.

We should add a "reaction" feature to a message that works similar to the "format" reaction feature, where if a user adds a "pin" feature, (the one that looks like a red tac) the bot takes that message, copies its contents, and DM's it to the user who pinned it, and then remove its reaction (so other users may pin the message or similar)

I'm not sure if the bot should wrap the original message in an embed, or just send it as it without any context.

**Additional context**
This should be done after the migration to TS #55

cjcanlas01:
- [x] Added the pinned message to an embed
- [x] Also added the author of the message as reference

Closes: #78

### DIFF
--- a/src/commands/bootstrap.ts
+++ b/src/commands/bootstrap.ts
@@ -53,6 +53,33 @@ export function bootstrap({
         console.error(error);
       }
     }
+
+    if (reaction.emoji.name === '📌') {
+      try {
+        reaction.users.fetch().then((users) => {
+          // Get user that reacted with the pushpin emoji
+          const user = users.first();
+          const pinnedEmbed = new MessageEmbed()
+            .setColor('#0099ff')
+            .setTitle("Here's your pinned message buddy!")
+            .addFields(
+              { name: 'Author', value: reaction.message.author },
+              { name: 'Content', value: reaction.message.content }
+            )
+            .setFooter('Happy Coding! 😁');
+
+          user?.send(pinnedEmbed);
+
+          // Remove reaction from the message
+          reaction.message.reactions.cache
+            .get('📌')
+            ?.remove()
+            .catch((err) => console.log('Failed to remove reaction: ', err));
+        });
+      } catch (error) {
+        console.error(error);
+      }
+    }
   });
   if (config.WELCOME_DM) {
     // we only send this command if the WELCOME_DM environment variable
--- a/src/commands/bootstrap.ts
+++ b/src/commands/bootstrap.ts
@@ -56,26 +56,22 @@ export function bootstrap({
 
     if (reaction.emoji.name === '📌') {
       try {
-        reaction.users.fetch().then((users) => {
-          // Get user that reacted with the pushpin emoji
-          const user = users.first();
-          const pinnedEmbed = new MessageEmbed()
-            .setColor('#0099ff')
-            .setTitle("Here's your pinned message buddy!")
-            .addFields(
-              { name: 'Author', value: reaction.message.author },
-              { name: 'Content', value: reaction.message.content }
-            )
-            .setFooter('Happy Coding! 😁');
+        const users = await reaction.users.fetch();
+        // Get user that reacted with the pushpin emoji
+        const user = users.first();
+        const pinnedEmbed = new MessageEmbed()
+          .setColor('#0099ff')
+          .setTitle("Here's your pinned message buddy!")
+          .addFields(
+            { name: 'Author', value: reaction.message.author },
+            { name: 'Content', value: reaction.message.content }
+          )
+          .setFooter('Happy Coding! 😁');
 
-          user?.send(pinnedEmbed);
+        user?.send(pinnedEmbed);
 
-          // Remove reaction from the message
-          reaction.message.reactions.cache
-            .get('📌')
-            ?.remove()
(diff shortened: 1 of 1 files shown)


src/commands/bootstrap.ts
```diff
@@ -53,6 +53,33 @@ export function bootstrap({
         console.error(error);
       }
     }
+
+    if (reaction.emoji.name === '📌') {
+      try {
+        reaction.users.fetch().then((users) => {
```
naomi-lgbt: Should be able to do this as an `await`, rather than having to use `.then`.

src/commands/bootstrap.ts
```diff
@@ -53,6 +53,33 @@ export function bootstrap({
...
+            )
+            .setFooter('Happy Coding! 😁');
+
+          user?.send(pinnedEmbed);
+
+          // Remove reaction from the message
+          reaction.message.reactions.cache
+            .get('📌')
+            ?.remove()
+            .catch((err) => console.log('Failed to remove reaction: ', err));
```
naomi-lgbt: Do we need this specific `.catch`? The `try/catch` block should already handle this, right?

naomi-lgbt (changes requested):

src/commands/bootstrap.ts
cjcanlas01: Have updated `.then` to `await`.

src/commands/bootstrap.ts
cjcanlas01: Yes, you are right. I have also updated this. :slightly_smiling_face:

bradtaniguchi (commented):
Code looks good to me, but I didn't test it yet (can't do it right now)

will probably check it out tomorrow, if it works then I'll approve them :)

naomi-lgbt (approved):
Looks good to me! Seems to work as intended. Only error I could get it to throw was on trying to pin an embed message.

bradtaniguchi:
> Only error I could get it to throw was on trying to pin an embed message.
So this means if a user tries to pin an embedded message, the bot explodes? 

I think this is because the bot will try to send the message as an embed itself, and just uses the `reaction.message.content` as-is, which probably doesn't work for an embedded message. We could look into what we can do with an embedded message to either:
- just not support pinning embedded message (which is kinda lame)
- "pin" the embedded message as-is in the user DM
- "un-embedd" if possible and send as normal text 

I believe the `Message` object has something on it saying if its an embed or not, so we should be able to know before having the bot decide what to do with it. IF anything we can just handle this error more elegantly and add support for pinning embedded messages later :)

src/commands/bootstrap.ts
```diff
@@ -53,6 +53,29 @@ export function bootstrap({
...
+      try {
+        const users = await reaction.users.fetch();
+        // Get user that reacted with the pushpin emoji
+        const user = users.first();
+        const pinnedEmbed = new MessageEmbed()
+          .setColor('#0099ff')
+          .setTitle("Here's your pinned message buddy!")
+          .addFields(
+            { name: 'Author', value: reaction.message.author },
+            { name: 'Content', value: reaction.message.content }
```
naomi-lgbt:
```suggestion
            { name: 'Content', value: reaction.message.content || "Embedded messages cannot be pinned."}
```
So I ran in to a similar issue with my personal bot (logging deleted message embeds). I created a fall-back value in case the `message.content` was empty (which is the case with embeds). Doing the same here should prevent the error.

naomi-lgbt (changes requested):

src/commands/bootstrap.ts
naomi-lgbt: This would mean we can't support pinning embeds - but that is a feature we could implement later on if we wanted to look in to it.

bradtaniguchi:
I haven't forgotten about this PR, just been busy and haven't been able to test :)

src/commands/bootstrap.ts
```diff
@@ -53,6 +53,29 @@ export function bootstrap({
         console.error(error);
       }
     }
+
+    if (reaction.emoji.name === '📌') {
```
naomi-lgbt, line 57:
```suggestion
    if (reaction.emoji.name === '📌') {
      try {
        const users = await reaction.users.fetch();
        // Get user that reacted with the pushpin emoji
        const user = users.first();
        const pinnedEmbed = new MessageEmbed()
          .setColor('#0099ff')
          .setTitle("Here's your pinned message buddy!")
          .addFields(
            { name: 'Author', value: reaction.message.author },
            {
              name: 'Content',
              value:
                reaction.message.content ||
                'Embedded messages cannot be pinned.'
            }
          )
          .setFooter('Happy Coding! 😁');

        user?.send(pinnedEmbed);

        // Remove reaction from the message
        reaction.message.reactions.cache.get('📌')?.remove();
      } catch (error) {
        console.error(error);
      }
    }
```
Should fix the linter error.

bradtaniguchi:
Will fix failing format errors in master. Too lazy to update the PR, as that requires me cloning it locally which I don't want to do.

PR #92, opened 2020-07-10
Feature/91/add trivia command

### LINKED ISSUE #91: Add "fun-fact"/trivia command
no

Have a new `fact` or `trivia` command that spits out some random facts. Similar to the quotes from the FCC site.

None.

**Additional context**
This **might** add to spam. Will need to think about it.

naomi-lgbt:
@bradtaniguchi I've thrown together a basic command that pulls from a separate file of trivia facts and sends an embed containing a random fact. Not sure if this is the most effective way to handle it. 

Also, now we need some random facts. :)

closes #91


src/APIs/trivia-facts.ts
```diff
@@ -0,0 +1,14 @@
...
+  },
+  {
+    fact: 'nhcarrigan is awesome',
+    source: 'Everyone!'
+  },
+  {
+    fact: 'bjorno might be evil',
+    source: 'bjorno'
+  }
+];
```
bradtaniguchi: Do you think we should make this a JSON file, rather than TS file. Its still up in the air of this is a good idea or not, but having it as JSON would make this a good target as an easy "beginner's" PR sorta file/feature.

naomi-lgbt: That's a good idea. I'll refactor that tomorrow. 🙂

naomi-lgbt:
I've reconfigured the trivia to a `JSON` file in the same location (we probably need a better name for the folder, tbh). Had to set the `resolveJsonModule` in the tsconfig to import the file.

bradtaniguchi:
So what do we want to do with this this? We can use quotes from the actual FCC repo, just like the site, rather than our own list of trivia facts? If we go that route, we can close this PR and take that approach instead.

naomi-lgbt:
I do still like the trivia idea, as well - could we include both? 
Have the trivia command for our local file of fun facts, and (as example) a `motivate` command that pulls from the quotes? 🤔

bradtaniguchi:
I think adding both should be fine for now. We can add the `motivate` command in another PR/issue :)

bradtaniguchi (approved):
LGTM 👍

Issue #90, opened 2020-07-09
Add support to manage previously suspended users if needed


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
Currently with the #79 feature, after a user is suspended, we save them to the database. After which we no longer have any visibility or information on who has been suspended.

**Describe the solution you'd like**
We could take multiple approaches to this problem
- Could provide another command to get who is suspended, which is limited by Discords character limits.
- Could provide a way for admins to get who is suspended by their name/id/search to get information on who has been suspended in the past
- Could provide general metrics on who has been suspended (like the number of people suspended)
- Could match/align suspended users with those under the suspended role.

**Additional context**
Should discuss what approach could and should be taken, before starting anything :)

naomi-lgbt:
I'd take the approach where an admin uses a `user <@username>` command and the bot returns an embed stating whether the user has been previously suspended or not.

This allows us to upscale - the command could be expanded to include the certification-based roles the user has, for example.

Issue #73, opened 2020-06-25
Add previous suspension feature


naomi-lgbt:
**Is your feature request related to a problem? Please describe.**
I received a request from Bjorno on Discord that we allow the suspend command to track if a user has previously been suspended. If the user has previously been suspended, the bot should continue with the command as normal (suspend user, create suspended channel, etc.) but should also DM the admin that calls the command to notify them of a prior suspension.

**Describe the solution you'd like**
I have some additional code prepared for this, but my solution requires a MongoDB connection and a Mongoose dependency. As the suspend command itself is currently still being developed, I thought it best to hold off on this feature until those kinks are worked out.

bradtaniguchi:
We can work on this one together after the migration to TS.

Did you use base mongodb, or mongoose? Once we integrate a database for the bot, it opens up a number of doors for other features. I can't think of any right now that we would want, but its something to think about. 

Also I'm going to add a new label for "db" related stuff

JoshuaPelealu:
Wait, can't it just use a simple `fs` function to create a new file of everyone that has been suspended before? Then use `fs` again to access it to check if they have the property or not.

naomi-lgbt:
I had used mongoose for this - which is compatible with TS and works well on my personal bot. 
Had to create interfaces for the Schemas though (as extensions to the Document).

PR #96, opened 2020-07-13
Feature/90/add user command

naomi-lgbt:
I've drafted up a user command, in relation to #90. The way I've configured it, the bot should return an embed containing the user's username, roles (in anticipation for certification-based roles), and previous suspension status. 

If the `MONGO_URI` isn't set, the bot won't try to find the suspension status at all. If it does not find a suspension status, it leaves that field off the embed. If the user that calls the command isn't a mod (based on `KICK_MEMBERS` permission), the bot won't include the previous suspension field. 

This can't be merged until #73 is done, and will likely need refactoring after that. But I'm creating the draft now so we can begin tweaking the way this command should function.


naomi-lgbt:
Closing - fresh PR made with rebased code

PR #98, opened 2020-07-16
Feat/95/abstract reactions

### LINKED ISSUE #95: Refactor how non-command features are handled to be "abstracted" away.
Right now there are a few "non-command" that work on "emojis". 

These commands should be abstracted away in a similar manor as the command-def interface. Together this should allow us to update the `help` command to include information for these commands.

**Additional context**
This is related to #94 but should be done at its own pace, as this requires code changes directly in the bot itself.

bradtaniguchi:
- refactor reaction commands, should allow us to more easily extend and document them along with normal commands.

Closes #95

**note** is a draft as I need to test it out myself, but your welcome to test it to make sure the reactions still work :)

### DIFF
--- a/src/reactions/format-reaction.ts
+++ b/src/reactions/format-reaction.ts
@@ -6,9 +6,9 @@ export const formatReaction: ReactionDef = {
   description: 'Guesses and formats the message',
   command: async (reaction) => {
     try {
-      await reaction.message.reactions.removeAll();
+      await addFormatting(reaction.message);
 
-      addFormatting(reaction.message);
+      reaction.message.reactions.cache.get('🤖')?.remove();
     } catch (error) {
       /* A common issue with this not working correctly, is
        * if the bot does not have permissions to remove reactions.
--- a/src/reactions/format-reaction.ts
+++ b/src/reactions/format-reaction.ts
@@ -10,9 +10,11 @@ export const formatReaction: ReactionDef = {
 
       addFormatting(reaction.message);
     } catch (error) {
-      // A common issue with this not working correctly, is
-      // if the bot does not have permissions to remove reactions.
-      // Check the README.md file for details on setting the correct permissions
+      /* A common issue with this not working correctly, is
+       * if the bot does not have permissions to remove reactions.
+       * Check the README.md file for details on setting the correct permissions
+       */
+
       reaction.message.channel.send(
         'Something went wrong! Failed to format code :('
       );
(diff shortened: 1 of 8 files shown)


naomi-lgbt:
I'll run this through some tests today too.

src/reactions/format-reaction.ts
```diff
@@ -0,0 +1,22 @@
...
+  description: 'Guesses and formats the message',
+  command: async (reaction) => {
+    try {
+      await reaction.message.reactions.removeAll();
+
+      addFormatting(reaction.message);
+    } catch (error) {
+      // A common issue with this not working correctly, is
+      // if the bot does not have permissions to remove reactions.
+      // Check the README.md file for details on setting the correct permissions
```
naomi-lgbt: Should this be changed to the multi-line comment format? To remain consistent with the rest of the code?

```suggestion
      /* A common issue with this not working correctly, is
       * if the bot does not have permissions to remove reactions.
       * Check the README.md file for details on setting the correct permissions
       */

```

src/reactions/pin.ts
```diff
@@ -0,0 +1,33 @@
...
+  emoji: '📌',
+  description: 'Sends message to the current user as a DM',
+  command: async (reaction) => {
+    try {
+      const users = await reaction.users.fetch();
+      // Get user that reacted with the pushpin emoji
+      const user = users.first();
+      const pinnedEmbed = new MessageEmbed()
+        .setColor('#0099ff')
+        .setTitle("Here's your pinned message buddy!")
```
naomi-lgbt: At some point I'd like to change this verbiage to something more professional...

naomi-lgbt (commented):
Tested and appears to function as intended. Have a couple of nit-picks but nothing feature-breaking.

src/reactions/pin.ts
bradtaniguchi: Got any ideas? We can change i now so we don't forget

bradtaniguchi:
I opened this up from draft.

I noticed that checking emojis by their name might not be the "safest" approach. On my system I can't even see the emojis correctly (probably due to something linux/system related) but it makes things a little more iffy when it comes to testing and verifying the code works. I'll probably refactor that part of the code and re-test later today.

@nhcarrigan let me know if you can think of a more professional title for the message later.

src/reactions/pin.ts
naomi-lgbt: Something like "Pinned Message:" would be sufficient, IMO.

bradtaniguchi:
My attempts at updating how emojis are checked didn't work so the PR is the same besides the 1 message change.

I noticed a somewhat important bug while testing, will create an issue for it

**edit** issue: #101

PR #103, opened 2020-07-17
Readme Script

### LINKED ISSUE #94: Create utility script prints out markdown specifications for commands
Right now the README lacks any information on what commands are available. 

 We should create a [markdown table](https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables) to display the list of commands. We can use the **same** command-def definitions from the files via a script that can be ran. 

This will allow us to keep the README in-line with the commands in a well formatted manor that is easy to upkeep.

**Additional context**
This should be done to support **most** of the command features available. Other commands/features like the "pin message" feature should be specified elsewhere in another table (this will be done later, once these commands are refactored)

**edit** it **may** be possible to run an automated check to make sure if all commands are included in the table in travis-ci, but I'm not sure if it could be used to actually keep it updated since it needs to be comited. We can look into this after this feature is actually working.

JoshQuaintance:
May close #94 
I have made a script that will create a string that will form a table into the README file. 

Since we are still debating about either auto Updating the README file or just `console.log()` the result, I made the script able to take a flag. If when running the script with the flag `--autoUpdateReadme`, it will auto update the readme, ex:
```bash
ts-node ./src/scripts/readmeScript.ts --autoUpdateReadme
```
The Command Line execution above will make the scrip auto update the README. But if the flag is not given, it will just `console.log()` the data table, ex:
```bash
ts-node ./src/scripts/readmeScript.ts
```
The Command Line execution above will make the script only `console.log()` the table results


src/scripts/readmeScript.ts
```diff
@@ -0,0 +1,35 @@
...
+// it will run this
+if (flags.includes('--autoUpdateReadme')) {
+  readFile('README.md', 'utf-8', (err, data) => {
+    if (err) throw err;
+
+    // Change the regex in this line to determine where you're going to put it at...
+    const results = data.replace(
+      /## Available Commands.*## Additional Information/gs,
+      commandTable
+    );
```
naomi-lgbt:
```suggestion
const commandTable = stripIndents`
  ## Available Commands
  | Prefix | Description |
  | :-: | :-: |
  ${COMMANDS.map(
    (command) => '| ' + command.prefix + ' | ' + command.description + ' |\n'
  ).join('')}
  ##
`;

// if there is an argument --autoUpdateReadme while running the command,
// it will run this
if (flags.includes('--autoUpdateReadme')) {
  readFile('README.md', 'utf-8', (err, data) => {
    if (err) throw err;

    // Change the regex in this line to determine where you're going to put it at...
    const results = data.replace(
      /## Available Commands[^#]*##/gs,
      commandTable
    );
```

The current default RegEx will remove the `Additional Information` header. Changing the code to this will have the script find the Available Commands header, and replace all content until it finds the next `##`.

naomi-lgbt (changes requested):

naomi-lgbt (approved):
LGTM

bradtaniguchi:
due to the merge with #104, manually fix the merge conflicts in the package.json, and one easy way to handle the merge conflicts in the package-lock.json is to delete it, and rerun `npm install`. There aren't that many conflicts so it could also be done manually.

JoshQuaintance:
Resolved Merge Conflicts! :)

package.json
```diff
@@ -13,6 +13,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "common-tags": "^1.8.0",
```
bradtaniguchi: move to devDependencies, as we don't need or want this installed when running the bot.

src/scripts/readmeScript.ts
```diff
@@ -0,0 +1,36 @@
+import { COMMANDS } from '../commands/commands';
+import { readFile, writeFile } from 'fs';
+import { stripIndents } from 'common-tags';
+
+const flags = process.argv.slice(2);
```
bradtaniguchi: document either here, or in the README on what flags are available.

src/scripts/readmeScript.ts
```diff
@@ -0,0 +1,36 @@
...
+  ${COMMANDS.map(
+    (command) => '| ' + command.prefix + ' | ' + command.description + ' |\n'
+  ).join('')}
+  ##
+`;
+
+// if there is an argument --autoUpdateReadme while running the command,
+// it will run this
+if (flags.includes('--autoUpdateReadme')) {
+  readFile('README.md', 'utf-8', (err, data) => {
```
bradtaniguchi: use `promise` based `readFile` + async/await instead of a callback

src/scripts/readmeScript.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (err) throw err;
+
+    // Change the regex in this line to determine where you're going to put it at...
+    const results = data.replace(
+      /## Available Commands[^#]*##/gs,
+      commandTable
+    );
+
+    writeFile('README.md', results, 'utf8', (err) => {
```
bradtaniguchi: use `promise` based `writeFile` + async/await instead of a callback

src/scripts/readmeScript.ts
```diff
@@ -0,0 +1,36 @@
+import { COMMANDS } from '../commands/commands';
+import { readFile, writeFile } from 'fs';
```
bradtaniguchi: Change to 
```ts
import { promises } from 'fs';
const { readFile, writeFile } = promises;
```

this will give you promise versions of the callback methods. Available after node 10 I believe (we are using 12, so it should be good)

bradtaniguchi (commented):
- rename `readmeScripts.ts` to `readme-scripts.ts`, to match with other files in repo
- move `scripts` folder outside of `src`, since this should be built/part of the bot
- add npm script to run command if possible

**note** running the script via an npm script will probably require another tsconfig.json that extends the top-level one, and maybe `ts-node` to execute it by itself.

JoshQuaintance:
Don't close this yet, might change the chalk dependency into something else...

bradtaniguchi:
@JoshuaPelealu `chalk` is pretty well respected, (it even comes with TS support!) so I'm fine with adding it as a dev-dep :)

JoshQuaintance:
Well, in that case, we can keep chalk as dev dep when we need it.

scripts/readme-script.ts
```diff
@@ -0,0 +1,54 @@
...
+        commandTable
+      );
+
+      writeFile('README.md', results, 'utf8')
+        .then(() => {
+          console.log('README.md updated');
+        })
+        .catch((err) => console.log(err));
+    })
+    .catch((err) => console.log(err));
```
bradtaniguchi: we **could** wrap the entire script into a `(async () => {})()` to get async/await, or if you want to use `.then` and `.catch`, this can be flattened like so:
```typescript
readFile('README.md', 'utf-8')
.then((data) => {
  // existing code
  return writeFile('README.md', results, 'utf8)
})
.then(() => console.log('README.md updated'))
.catch(err => console.error(err));
```

Prettier may format that differently, but then taking this approach prevents the "callback hell" structuring that is currently happening with the promises.

scripts/readme-script.ts
```diff
@@ -0,0 +1,54 @@
...
+if (flags.includes('--autoUpdateReadme')) {
+  readFile('README.md', 'utf-8')
+    .then((data) => {
+      // Change the regex in this line to determine where you're going to put it at...
+      const results = data.replace(
+        /## Available Commands.*## Additional Information/gs,
+        commandTable
+      );
+
+      writeFile('README.md', results, 'utf8')
```
bradtaniguchi: utf8 is different than the "readFile` call of `utf-8` is that ok? Or just inconsistent

JoshQuaintance: That is ok since javascript allows dashes `-`. The `utf8` is usually used for programs that doesn't really support dashes. It's just inconsistent in my end. It will both tell the command that the format is `utf-8`

JoshQuaintance: I will just edit it anyway

bradtaniguchi (approved):
LGTM :+1:

PR #107, opened 2020-07-20
Feature/90/add user command v2

### LINKED ISSUE #90: Add support to manage previously suspended users if needed

naomi-lgbt:
# Pull Request

## Description:
I've drafted up a user command, in relation to #90. The way I've configured it, the bot should return an embed containing the user's username, roles (in anticipation for certification-based roles), and previous suspension status. 

If the `MONGO_URI` isn't set, the bot won't try to find the suspension status at all. If it does not find a suspension status, it leaves that field off the embed. If the user that calls the command isn't a mod (based on `KICK_MEMBERS` permission), the bot won't include the previous suspension field. 

## Related Issue:

Closes #90 
Replaces PR #96 (too many merge conflicts)


bradtaniguchi (commented):
Shouldn't this command be limited to admins?

Also fix the merge conflict :)

naomi-lgbt:
I wrote the command such that the `suspended` field was only generated when an admin called the command, but that the roles and username fields generate for anyone who calls the command. My thought behind this was that when certification-based roles become a thing, this will allow users to quickly see which certifications another user has completed (so they can ask for help) or show off their own certifications. 🙂

bradtaniguchi:
Ah I didn't see that part of the command as I didn't look closely enough. Discord already shows what roles people have, so this might not be needed. I thought the purpose of this command was just to show who was suspended before, and not so much show their roles.

naomi-lgbt:
I could refactor it to only offer the suspended information - I only went this route in anticipation of potential features added in the future. 

For example, we could also add a field to the embed that shows when a user joined the server.

bradtaniguchi:
That makes sense as a feature. I think the main goal of this feature is providing more metadata on users that would be helpful **from an admin standpoint**. We can expand on this feature in another PR/issue after we look into what would be useful information, and what can we do. 

For now just supporting anything related to suspended, and locking this down to just admins should be good enough. We can open another issue with ideas/concepts for what expanding on this looks like :)

naomi-lgbt:
Done and done :)

src/commands/user.ts
```diff
@@ -0,0 +1,35 @@
...
+      console.log('No database configured');
+      return;
+    }
+    const userSuspend: UserSuspend | null = await userSuspendModel.findOne({
+      userId: user.id
+    });
+    let suspended = userSuspend?.suspended;
+    if (!suspended) {
+      suspended = false;
+    }
```
bradtaniguchi: This can be done with just:
```suggestion
    const suspended = !!userSuspend?.suspended;
```

bradtaniguchi (commented):
Change one thing, otherwise LGTM :+1:

PR #109, opened 2020-07-20
Fix/deleted message logging

naomi-lgbt:
# Pull Request

## Description:
Assuming the bot is configured to log deleted messages, when a message containing an embed is deleted the bot explodes. The source of the error is that `message.content` is undefined for an embed - this causes the log embed to have an empty value, which throws an error. I've added a fallback value (much like the pinned message command) to prevent this from occuring. 
## Related Issue:

n/a


bradtaniguchi (approved):
LGTM :+1:

PR #111, opened 2020-07-21
Feature/add quote command

naomi-lgbt:
# Pull Request

## Description:
Adds a quote command that pulls from the FreeCodeCamp motivational quotes. Returns an embed with the the title as a compliment from the file, the main content as the quote, and the footer as the quote author.
## Related Issue:

Discussed in #91


src/commands/quote.ts
```diff
@@ -0,0 +1,23 @@
...
+import { MessageEmbed, Message } from 'discord.js';
+
+export const quote: CommandDef = {
+  prefix: 'quote',
+  description: 'Returns a quote from the FCC motivational quotes file',
+  command: async (message) => {
+    const data = await fetch(
+      'https://raw.githubusercontent.com/freeCodeCamp/freeCodeCamp/master/config/motivational-quotes.json'
+    );
+    const parsed = await data.json();
```
bradtaniguchi: To prevent spamming github, we should do this **only 1 time when the bot starts**. We can then pass the data thru the 2nd argument of the command. (name this `quotesDefs` as a param, with a type-def reflecting the structure of the quotes.)

naomi-lgbt: Okay, I've done this - I feel like I made it more complicated than I needed to?? Let me know if I didn't quite hit the mark on this.

src/APIs/quote-def.ts
```diff
@@ -0,0 +1,4 @@
+export interface quoteDef {
```
bradtaniguchi: Name interfaces similar to `class` declarations.

```suggestion
export interface QuoteDef {
```

src/APIs/quote-def.ts
```diff
@@ -0,0 +1,4 @@
...
+  compliments: Array<string>;
+  motivationalQuotes: Array<Record<string, unknown>>;
```
bradtaniguchi: This should be defined differently:

```suggestion
  motivationalQuotes: Array<{
    quote: string;
    author: string;
  }>;
```

we **could** define a `MotivationalQuote` interface, but the "inline" one should be fine for now.

bradtaniguchi (commented):
Update the TS interface and the code LGTM :+1:

( I haven't tested it yet tho)

bradtaniguchi (approved):
LGTM :+1:

Issue #119, opened 2020-07-31
[FEATURE/CHORE] Update console log statements to a logger


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
Right now all logs are handled by `console.log` or similar.

**Describe the solution you'd like**
We should migrate all of these calls to using something like [winston](https://github.com/winstonjs/winston). This will allow us to be more flexible while logging and getting more context. (such as information on who did what)

**Describe alternatives you've considered**
We can stick with our current logging system, but it doesn't scale or give enough context often.

**Additional context**
This process should be done slowly, and can be spread across multiple PR requests.

JoshQuaintance:
I will start by installing winston as a dependency, then I will go through the docs see what I can do.

JoshQuaintance:
So, I've been tinkering around with winston, and it seems like we can make the logger log in a json format, or in a format that we can configure ourselves.

naomi-lgbt:
@JoshuaPelealu Where are you at with this?

@bradtaniguchi I have been thinking - at least some of these console logs should probably be migrated to end-user messages. Otherwise we face the struggle of the bot appearing to be unresponsive, instead of providing feedback to the user about why the command didn't work.

bradtaniguchi:
@nhcarrigan I think we can update some error states to generic end user messages, **and** still log with more extensive logging information.


Also, I'll start looking into this :)

PR #121, opened 2020-08-02
Feature testing

### LINKED ISSUE #120: [FEATURE] Update Help command to be able to take parameters and a nicer help message
Right now the help command only sends a message to the user and send the channel a message notifying that a help message was sent.

We can improve this so that the command will be able to take a parameter to a command, and reply with the command name, description, and how to use it. We can also give every command a parameter where we can specify how to use each command.

We can at least make the help command instead of sending such a long list with description, just send a comma seperated list of all the commands into the channel the user is using it at. Then they can use the `help [command]` feature to find out more about each command

**Additional context**
This will make the help statemenet for the bot be a lot cleaner and nicer, that the user can use easily.

JoshQuaintance:
**Description:**
The help command now take parameters. If it doesn't recieve any, it will send in a list of all the commands. If the command receives a valid command, it will find it in the command array, and then send the user the usage and the description for the command

Closes #120 

- [x] Takes parameter
- [x] Shows list of all commands if param is not specified
- [x] Sends the usage and description for commands
- [x] Added a new value to each command: `usage` that will be used to display usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "prestart": "npm run build",
     "start": "node dist",
     "test": "jest --detectOpenHandles",
     "dev": "npx nodemon --inspect",
@@ -23,8 +22,7 @@
     "mongoose": "^5.9.20",
     "node-fetch": "^2.6.0",
     "prettier": "^2.0.5",
-    "strip-ansi": "^6.0.0",
-    "winston": "^3.3.3"
+    "strip-ansi": "^6.0.0"
   },
   "devDependencies": {
     "@types/common-tags": "^1.8.0",
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
+    "prestart": "npm run build",
     "start": "node dist",
     "test": "jest --detectOpenHandles",
     "dev": "npx nodemon --inspect",
@@ -22,7 +23,8 @@
     "mongoose": "^5.9.20",
     "node-fetch": "^2.6.0",
     "prettier": "^2.0.5",
-    "strip-ansi": "^6.0.0"
+    "strip-ansi": "^6.0.0",
+    "winston": "^3.3.3"
   },
   "devDependencies": {
     "@types/common-tags": "^1.8.0",
--- a/src/commands/close.ts
+++ b/src/commands/close.ts
@@ -7,6 +7,7 @@ export const closeCommand: CommandDef = {
     'Closes the channel. This command requires admin privileges, ' +
     'and will only work on the automatically created "suspended" channels.' +
     ' Include the user if you want to remove the suspended role.',
+  usage: 'close',
   command: async (message, { config }): Promise<void> => {
     try {
       const target = message.channel as TextChannel;
--- a/src/commands/close.ts
+++ b/src/commands/close.ts
@@ -1,13 +1,15 @@
 import { CommandDef } from './command-def';
 import { TextChannel, MessageEmbed } from 'discord.js';
+import { oneLine } from 'common-tags';
 
 export const closeCommand: CommandDef = {
   prefix: 'close',
-  description:
-    'Closes the channel. This command requires admin privileges, ' +
-    'and will only work on the automatically created "suspended" channels.' +
-    ' Include the user if you want to remove the suspended role.',
-  usage: 'close',
+  description: oneLine`
+    Closes the channel. This command requires admin privileges,
+    and will only work on the automatically created "suspended" channels.
+    Mentioning user with the command will remove the suspended role from the user.
+  `,
+  usage: 'close [user]',
   command: async (message, { config }): Promise<void> => {
     try {
       const target = message.channel as TextChannel;
@@ -16,21 +18,18 @@ export const closeCommand: CommandDef = {
         (channel) => channel.name === config.LOG_MSG_CHANNEL
       ) as TextChannel;
       //check for user permissions
-      if (!message.member?.hasPermission('MANAGE_CHANNELS')) {
-        console.log('Missing permissions.');
-        return;
-      }
-      if (!log) {
-        console.log('Log channel not found.');
-        return;
-      }
+      if (!message.member?.hasPermission('MANAGE_CHANNELS'))
+        return console.log('Missing permissions.');
+
+      if (!log) return console.log('Log channel not found.');
+
       if (
         !target.name.includes('suspended') ||
         !target.parent?.name.includes(config.SUSPEND_CATEGORY)
       ) {
-        console.log('Cannot delete non-temporary channel');
-        return;
+        return console.log('Cannot delete non-temporary channel');
       }
+
       let status = 'The appeal was not approved.';
       if (message.mentions.members?.first()) {
         const suspend = message.guild?.roles.cache.find(
--- a/src/commands/coc.ts
+++ b/src/commands/coc.ts
@@ -7,7 +7,7 @@ export const coc: CommandDef = {
   usage: 'coc',
   command: (message): void => {
     const codeEmbed = new MessageEmbed()
-      .setTitle('FreeCodeCamp Code of Conduct')
+      .setTitle('freeCodeCamp Code of Conduct')
       .setDescription(
         'These are the basic rules for interacting with the FreeCodeCamp community on any platform, including this Discord server. You can read the full document on the [FreeCodeCamp article](https://freecodecamp.org/news/code-of-conduct)'
       )
@@ -27,7 +27,8 @@ export const coc: CommandDef = {
           value:
             'Spamming includes posting off-topic messages to disrupt discussions, promoting a product, soliciting donations, advertising a job / internship / gig, or flooding discussions with files or text.'
         }
-      );
+      )
+      .setFooter("Thank you for following freeCodeCamp's Code of Conduct");
 
     message.channel.send(codeEmbed).catch((error) => console.log(error));
   }
--- a/src/commands/coc.ts
+++ b/src/commands/coc.ts
@@ -4,6 +4,7 @@ import { CommandDef } from './command-def';
 export const coc: CommandDef = {
   prefix: 'coc',
   description: 'Provides the Code of Conduct.',
+  usage: 'coc',
   command: (message): void => {
     const codeEmbed = new MessageEmbed()
       .setTitle('FreeCodeCamp Code of Conduct')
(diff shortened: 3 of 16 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
+    "prestart": "npm run build",
```
bradtaniguchi: How come this was added?

package.json
```diff
@@ -22,7 +23,8 @@
     "mongoose": "^5.9.20",
     "node-fetch": "^2.6.0",
     "prettier": "^2.0.5",
-    "strip-ansi": "^6.0.0"
+    "strip-ansi": "^6.0.0",
+    "winston": "^3.3.3"
```
bradtaniguchi: I think you were also testing out changes for #119 and this got mixed in? 
You can revert this change by uninstalling this package, and or reverting the package.json and package-lock.json files.

src/commands/close.ts
```diff
@@ -7,6 +7,7 @@ export const closeCommand: CommandDef = {
     'Closes the channel. This command requires admin privileges, ' +
     'and will only work on the automatically created "suspended" channels.' +
     ' Include the user if you want to remove the suspended role.',
+  usage: 'close',
```
bradtaniguchi: I think this command actually takes an argument for a "reason"? Might need to check with @nhcarrigan since he wrote this and I don't remember off the top of my head. (I also am being lazy right now and don't want to go thru the code, sorry :'( )

src/commands/eightball.ts
```diff
@@ -5,6 +5,7 @@ export const eightBall: CommandDef = {
   prefix: 'eightball',
   description:
     'Think of your question and get an answer from a magic pool ball!',
+  usage: `\`eightball [message]\``,
```
bradtaniguchi: This is incorrect, this command doesn't take a message. This might of been done first and just forgotten about.
Please change to:
```suggestion
  usage: 'eightball',
```

src/commands/help.ts
```diff
@@ -1,37 +1,63 @@
...
+      if (!commandFound)
+        return message.channel.send(
+          `Cannot find the command you specified -> \`${param}\``
+        );
+      const helpEmbed = new MessageEmbed()
+        .setColor('#0099FF')
+        .setTitle(commandFound.usage)
+        .addField('Description', `\`${commandFound.description}\``)
+        .addField('\u200b', 'Required: `<>` | Optional: `[]`');
+      message.channel.send(helpEmbed);
```
bradtaniguchi: The bot should send messages to the user in a DM rather than in chat, to prevent spam.

src/commands/help.ts
```diff
@@ -1,37 +1,63 @@
...
+          'Bot Information',
+          'Hello! I am a test bot created by bradtaniguchi and members of ' +
+            'FreeCodeCamp.org to experiment with the process of building a ' +
+            'Discord Bot. You can view my [source code here](https://github.com/bradtaniguchi/discord-bot-test)'
+        )
+        // .addField(
+        //   'Commands',
+        //   `Hey there ${message.author}, here is the full list of all my commands.\n` +
+        //     `For more information about each command, use \`${config.PREFIX} help <command>\``
+        // )
```
bradtaniguchi: remove comments

src/commands/help.ts
```diff
@@ -1,37 +1,63 @@
...
-        name: commandDef.prefix,
-        value: commandDef.description.substring(0, 1023),
-        inline: false
-      });
-    });
-    await message.author.send({ embed: helpEmbed });
-    message.channel.send('Help message sent!');
+  command: async (message, { config }) => {
+    const param = message.content.split(' ')[2];
+    if (param == undefined) {
```
bradtaniguchi: - Change `param` to `command`
- Change `if (param == undefined)` to `if(!param)` and **move the logic from the else to this if**.
- Move the logic in this command function to 2 separate functions, rather than if/else. Pass along any extra params in a single argument (similar to how the 2nd argument of the `command-def` works)

package.json
JoshQuaintance: Owh, I added that myself, I just forgot to remove it again. It's just for my convinience so it will always  build before starting the bot. But I'll remove it.

src/commands/close.ts
JoshQuaintance: Alright, I'll make sure to update it.

naomi-lgbt: Correct - the `close` command takes an optional `user` parameter (in the form of an @ mention). If the parameter is provided, that user is unsuspended.

src/commands/user.ts
```diff
@@ -4,7 +4,8 @@ import { MessageEmbed } from 'discord.js';
 
 export const userCommand: CommandDef = {
   prefix: 'user',
-  description: "Get data on a current user. Use the syntax 'user [username]'.",
+  description: 'Get data on a given user, or the command caller themselves.',
```
naomi-lgbt: This is incorrect - the `user` command will not work if a user is not provided. 
This was done intentionally, as the `user` command is currently intended to tell a moderator if a user has previously been suspended. There is no need for a moderator to check this information on their own username :)

src/commands/suspend.ts
```diff
@@ -5,8 +5,8 @@ import { userSuspendModel, UserSuspend } from '../APIs/mongo-suspend';
 export const suspendCommand: CommandDef = {
   prefix: 'suspend',
   description:
-    'Suspends a user for the given reason. This command is only available to admins. ' +
-    'Use the format "suspend <usertag> <reason>"',
+    'Suspends a user for the given reason. This command is only available to admins. ',
+  usage: 'suspend <usertag> <reason>',
```
naomi-lgbt: The `reason` parameter is optional - the bot will use "No reason provided" if this parameter is void.

src/commands/help.ts
naomi-lgbt: The original command sent the embed directly to the author, and a "Help message sent" confirmation to the channel (to prevent the bot appearing to be unresponsive to other users).

naomi-lgbt (changes requested):

package.json
bradtaniguchi: Just to check, during development you can run `npm run dev`, which will run and live-reload the app. :)

src/commands/user.ts
JoshQuaintance: Owh, so this is not like the user command that gives information about the user.

JoshQuaintance: Fixing it

naomi-lgbt: Yeh - it may be refactored later, but currently it only gets the prior suspension data.

src/commands/suspend.ts
JoshQuaintance: Fixed

src/commands/coc.ts
```diff
@@ -4,9 +4,10 @@ import { CommandDef } from './command-def';
 export const coc: CommandDef = {
   prefix: 'coc',
   description: 'Provides the Code of Conduct.',
+  usage: 'coc',
   command: (message): void => {
     const codeEmbed = new MessageEmbed()
-      .setTitle('FreeCodeCamp Code of Conduct')
+      .setTitle('freeCodeCamp Code of Conduct')
```
naomi-lgbt, line 10: Good catch. XD

src/commands/help.ts
```diff
@@ -1,37 +1,88 @@
...
-    });
-    await message.author.send({ embed: helpEmbed });
-    message.channel.send('Help message sent!');
+  command: async (message, { config }) => {
+    const Command = message.content.split(' ')[2];
+    if (!Command) {
+      sendFullList({ message, config });
+    } else {
+      sendSpecificCommand({ message, Command });
+    }
```
naomi-lgbt: Can leverage early returns to remove the `else`. @bradtaniguchi hates `else` XD

src/commands/quote.ts
```diff
@@ -4,7 +4,9 @@ import { MotivationalDef } from '../APIs/quote-def';
 
 export const quote: CommandDef = {
   prefix: 'quote',
-  description: 'Returns a quote from the FCC motivational quotes file',
+  description:
+    "Returns a quote fetched from freeCodeCamp's motivational quotes API",
```
naomi-lgbt: To clarify, it's fetched from the FCC `motivational-quotes.json` file, but the fetch uses the **GitHub** API. :)

src/commands/user.ts
```diff
@@ -4,21 +4,18 @@ import { MessageEmbed } from 'discord.js';
 
 export const userCommand: CommandDef = {
   prefix: 'user',
-  description: "Get data on a current user. Use the syntax 'user [username]'.",
+  description: 'Get Data on the user mentioned.',
```
naomi-lgbt: Why is `Data` capitalised here? 🤔

src/commands/help.ts
```diff
@@ -1,37 +1,88 @@
...
+    );
+  const helpEmbed = new MessageEmbed()
+    .setColor('#0099FF')
+    .setTitle(`\`${commandFound.usage}\``)
+    .addField('Description', commandFound.description)
+    .addField('\u200b', 'Required: `<>` | Optional: `[]`');
+
+  message.author.send(helpEmbed);
+  message.channel.send(
+    'Your help message with the command you requested is sent through DM'
```
naomi-lgbt:
```suggestion
    `DM sent with information on the ${commandFound.prefix} command`
```

Slight wording change.

src/commands/help.ts
```diff
@@ -1,37 +1,88 @@
...
+        stripIndent`
+        \u200b
+        \u200b
+        For more information about each command, use \`${
+          (config as Config).PREFIX
+        } help <command>\``
+    )
+    .setFooter('I am not affiliated with FreeCodeCamp in any way.');
+  message.author.send(helpEmbed);
+  message.channel.send('Your help message is sent through DM');
```
naomi-lgbt:
```suggestion
  message.channel.send('DM sent with the requested information');
```

Slight wording change. Keeps it a little cleaner than using the second-person perspective.

naomi-lgbt (changes requested):
Some more thoughts I had.

src/commands/user.ts
JoshQuaintance: I didn't even know.

src/commands/eightball.ts
```diff
@@ -5,6 +5,7 @@ export const eightBall: CommandDef = {
   prefix: 'eightball',
   description:
     'Think of your question and get an answer from a magic pool ball!',
+  usage: `eightball [message]`,
```
naomi-lgbt: Last one - remove this parameter and then LGTM

naomi-lgbt (changes requested):

naomi-lgbt (approved):
LGTM!

PR #122, opened 2020-08-05
Feature/add mod command

naomi-lgbt:
**Description:**

Adds a command to provide helpful links for moderators - though the links are available to anyone, so the command is not role-restricted.


src/commands/mod.ts
```diff
@@ -0,0 +1,29 @@
...
+        value: 'https://contribute.freecodecamp.org/'
+      },
+      {
+        name: 'News Contributing',
+        value: 'https://www.freecodecamp.org/news/developer-news-style-guide/'
+      }
+    );
+    message.channel.send(modEmbed);
+  },
+  usage: `mod`
```
bradtaniguchi: Keep this at the top under description. To match with the other commands, and so it doesn't get "lost" at the end of the command function.

bradtaniguchi (approved):
Make 1 change, and then after that LGTM :+1:

naomi-lgbt:
Fixed 🙂

PR #123, opened 2020-08-05
Fix/dependencies

naomi-lgbt:
**Description:**

The heroku instance won't launch - the `common-tags` package is listed as a dev-dependency, which heroku clears after build. There are commands that rely on this dependency, so I've moved it to production.

Closes #XXXXX


bradtaniguchi (approved):
LGTM :+1: but **really** look into updating your fork so there aren't all these extra merge conflicts :)

PR #131, opened 2020-08-12
Fix/unlock suspend features

naomi-lgbt:
**Description:**

This PR turns on the automatic channel creation and message for the `suspend` command.

Closes #XXXXX


bradtaniguchi (approved):
didn't test anything, but LGTM. If there are problems we can handle em :D

Issue #135, opened 2020-08-12
[FEATURE] Update suspend command to provide more resources and information


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
Currently the `suspend` command provides the following:
![image](https://user-images.githubusercontent.com/10079147/90065014-75f3a000-dca0-11ea-9435-9dda20665e7e.png)
We should provide more information in this message, to the user for their suspending, as the bot is a "third party" it can act as a moderator to make the user more aware of the situation rather than the traditional DM from the actual mod who performed the suspension.

**Describe the solution you'd like**
We should provide the actual moderation template instead. (as suggested by @nhcarrigan)

**Describe alternatives you've considered**
We could also add more "sub-commands" available to the bot within a `suspended channel` if need be in the future :)

**Additional context**
TODO: add moderation template link here, or hopefully someone else does below :)

naomi-lgbt:
This should be good now, with the merge of #138?

bradtaniguchi:
Yea we can close this for now

PR #138, opened 2020-08-12
Feature/revamp suspend command

naomi-lgbt:
**Description:**
This PR tackles a few things:

Suspend command:
1. Explicitly deny the `create invite` permission for everyone in the automatically created channel. 
2. Remove the DM sent to the user on suspension. 
3. Use the FCC template for the initial message in the suspended channel.
4. Include the username and the reason in that message.
5. That message also tags the moderator role. 

Close command:
1. Now requires an additional parameter of "accepted" or "denied". Will fail on any other value. 
2. "Accepted" closes the channel and removes the suspension. "Denied" only closes the channel.
3. Now requires the `manage channels` permission. 

TODO:
~~1. Close command should require "manage channels" permission~~
~~2. Close command should take "accepted" | "denied" to determine whether to remove suspension or not.~~

### DIFF
--- a/src/commands/close.ts
+++ b/src/commands/close.ts
@@ -12,6 +12,10 @@ export const closeCommand: CommandDef = {
   usage: 'close [accepted | denied] [user]',
   command: async (message, { config }): Promise<void> => {
     try {
+      if (!message.member?.hasPermission('MANAGE_CHANNELS'))
+        return console.log(
+          `${message.author.username} did not have the correct permissions.`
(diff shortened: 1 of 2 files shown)


bradtaniguchi:
Related to #139 (these are the TODOs)

naomi-lgbt:
Also addresses #135.

bradtaniguchi (approved):
LGTM :+1:

PR #142, opened 2020-08-15
Fix/forum date parsing

naomi-lgbt:
**Description:**

Fixes the date format for the `forum` command. Readable dates are nice.

Closes #XXXXX


src/commands/forum.ts
```diff
@@ -15,10 +15,12 @@ export const forum: CommandDef = {
       const forumEmbed: MessageEmbed = new MessageEmbed()
         .setTitle('Latest Forum Activity')
         .setDescription('Here are the five most recent posts.');
-      result.forEach((el: Record<string, unknown>) =>
+      result.forEach((el: Topic) =>
```
bradtaniguchi: The `result` variable can be renamed to something like `topics` and typed on the `.slice(0, 5)`  with the type `Topic[]`.

bradtaniguchi (commented):
make 1 change so the code is more "inferred" then LGTM :+1:

bradtaniguchi (approved):
LGTM :+1:

PR #151, opened 2020-08-18
Leave message only on suspended role

naomi-lgbt:
**Description:**

This PR modifies the "user left" logging to only send a message if the user is suspended when they leave.

Closes #XXXXX

### DIFF
--- a/src/commands/bootstrap-commands.ts
+++ b/src/commands/bootstrap-commands.ts
@@ -56,9 +56,11 @@ export const bootstrapCommands = ({
       if (goodbyeChannel.type !== 'text') {
         console.error('log ');
       }
-      (goodbyeChannel as TextChannel).send(
-        `${member.user} (**${member.user?.username}**) has left us! :(`
-      );
+      if (member.roles.cache.has('720740563472744521')) {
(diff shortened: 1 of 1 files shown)


src/commands/bootstrap-commands.ts
```diff
@@ -56,9 +56,11 @@ export const bootstrapCommands = ({
       if (goodbyeChannel.type !== 'text') {
         console.error('log ');
       }
-      (goodbyeChannel as TextChannel).send(
-        `${member.user} (**${member.user?.username}**) has left us! :(`
-      );
+      if (member.roles.cache.has('720740563472744521')) {
```
bradtaniguchi: shouldn't this be checking against the `SUSPENDED_ROLE` [environment variable](https://github.com/bradtaniguchi/discord-bot-test/blob/master/src/config/get-config.ts#L58)?

I dislike "magic strings/numbers"

naomi-lgbt: Yeh that's a better approach. I'll refactor it :)

bradtaniguchi (approved):
LGTM :+1:

PR #153, opened 2020-08-19
Feature/suspend logging

### LINKED ISSUE #152: [FEATURE/DISCUSSION] Start working on "userlog" feature for admins
Currently moderation on the discord server doesn't provide any way to make "notes" or save references about a user's behavior. 

Add a new `userlog` command based off the following examples:
```
fCC userlog @bradtaniguchi User was suspended for advertising (bot saves link to this message)
fCC userlog @bradtaniguchi User was suspended again, this time for disruptive behavior (bot saves link to this message)
```
so:
```
fCC userlog <user> <...rich text message>
```

This will require updates to the database to save the `userId` mentioned, **and** their current username, along with other meta data. (we can talk about the database schema) Any discord links should be saved as-is to be displayed later

We could add another command that can be used to "read" this information later, or use the same command without the third argument. Regardless of how the "read" is called, it should show these logs, previous suspensions and their username (as these can be changed)

**Additional context**
This feature will require the usage of the mongodb database, and probably some database updates overall so user's are changed.

naomi-lgbt:
**Description:**

It looks like we had a lot of the framework for this one already, but it just needed some tweaking. Here's the workflow this PR establishes: 

1. Mod calls the `suspend` command on a user (and optionally provides reason).
2. Bot suspends user like normal.
3. Bot checks database for existing user DB entry (by userID, as AFAIK that is immutable), and pushes a `{date, reason}` object to the `suspended` array if it finds the user. If it does not find the user, it creates a new record for them.
4. Mod can call `user` command on any mention to get the user's suspension history. If the bot cannot find the userID in the DB, it responds with "clean record". 

This *should* cover all of Bjorno's asks for this one. If we choose to go live, I can slap my mongo URI in the Heroku env.

Closes #152


src/commands/user.ts
```diff
@@ -22,9 +22,11 @@ export const userCommand: CommandDef = {
...
     }
-    const log = userSuspend.suspended
-      .map((el) => `${el.date}: ${el.reason}`)
-      .join(' | ');
-    message.channel.send(`Here is the record for ${user}: ${log}`);
+    message.channel.send(`Generating suspend record for ${user}...`);
+    userSuspend.suspended.forEach((el) => {
+      message.channel.send(
+        `Suspended on ${el.date.split(',')[0]} for ${el.reason}`
+      );
```
bradtaniguchi: We want to send separate messages? I was thinking of just adding `\n` for each thing?

naomi-lgbt: We could go that route, too.

naomi-lgbt:
Okay @bradtaniguchi I have further modified it. 

The log now includes the mod that performs the suspension. The user schema tracks the user's current user name and nickname, and provides them when the `user` command is called. Ultimately it's all tied to the `userId` as that doesn't change, but this should work well enough I think?

bradtaniguchi:
I was thinking about de-normalizing the database data into different schemas based around the idea that we can save **other** stuff against the same user rather than shoving everything into 1 schema as we have now.

Might have to think about it a little more tho, I'm unsure about having everything in 1 schema, it might be fine, or might not be.

naomi-lgbt:
So you're thinking have a User schema for the username and nickname, and a Suspend schema for the log? And tie them together by the userID?

bradtaniguchi:
Yes **or** we de-normalize everything in 1 schema, but put everything in arrays, as there shouldn't be many items per user so there isn't much point in putting it into different schemas.

I can create an example of how the data will be saved in the db later if you want

naomi-lgbt:
@bradtaniguchi okay I switched it to target a new branch on the base repo.

PR #157, opened 2020-08-26
Feat/152/suspend user log

bradtaniguchi:
**Description:**
- Refactor schema/model setup with mongoose to reflect the new "userlog" feature, while making some changes to support pre-existing suspend logic

Related to #152 

**note** this is a draft since I want to get some feedback on how the schema is currently setup. I'm also not sure about how the `userId` isn't the same as the mongodb `_id` property, which could just be made to be the discord `snowflake` identifier instead. (its a rather long unique number)

### DIFF
--- a/src/db/moderator.ts
+++ b/src/db/moderator.ts
@@ -18,10 +18,6 @@ export interface Moderator {
    * The discord server nickname of the mod
    */
   username: string;
-  /**
-   * The discord server nickname of the mod
-   */
-  nickname: string;
 }
 
 export const moderatorSchema = new Schema({
@@ -34,12 +30,6 @@ export const moderatorSchema = new Schema({
     // See snowflake documentation: https://discord.com/developers/docs/reference#snowflakes
   },
   username: {
-    type: Schema.Types.String,
-    required: true,
-    immutable: true
-    // TODO: Add entry limitations, check discord api
-  },
-  nickname: {
     type: Schema.Types.String,
     required: true,
     immutable: true
--- a/src/db/moderator.ts
+++ b/src/db/moderator.ts
@@ -11,9 +11,9 @@ import { Schema } from 'mongoose';
  */
 export interface Moderator {
   /**
-   * The discord server userId for the mod
+   * The discord user's _id
    */
-  userId: string | Snowflake;
+  _id: Snowflake;
   /**
    * The discord server nickname of the mod
    */
@@ -24,26 +24,25 @@ export interface Moderator {
   nickname: string;
 }
 
-export const moderatorSchema = new Schema(
-  {
-    userId: {
-      type: Schema.Types.String,
-      required: true,
-      immutable: true
-      // TODO: Add length validation check
-    },
-    username: {
-      type: Schema.Types.String,
-      required: true,
-      immutable: true
-      // TODO: Add entry limitations, check discord api
-    },
-    nickname: {
-      type: Schema.Types.String,
-      required: true,
-      immutable: true
-      // TODO: Add entry limitations, check discord api
-    }
+export const moderatorSchema = new Schema({
+  _id: {
+    type: Schema.Types.String,
+    required: true,
+    immutable: true,
+    unique: true
+    // TODO: Add length validation check, is it 17 or 18???
+    // See snowflake documentation: https://discord.com/developers/docs/reference#snowflakes
   },
-  { _id: false }
-);
+  username: {
+    type: Schema.Types.String,
+    required: true,
+    immutable: true
+    // TODO: Add entry limitations, check discord api
+  },
+  nickname: {
+    type: Schema.Types.String,
+    required: true,
+    immutable: true
+    // TODO: Add entry limitations, check discord api
+  }
+});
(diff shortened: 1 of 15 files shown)


src/db/moderator.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  userId: string | Snowflake;
+  /**
+   * The discord server nickname of the mod
+   */
+  username: string;
+  /**
+   * The discord server nickname of the mod
+   */
+  nickname: string;
```
naomi-lgbt: To clarify, the nickname can be `null`. You and I both do not have nicknames set, so this field is empty. To typecast as a string, I used the `member.nickname` field and a fallback of `""`.

src/db/user.ts
```diff
@@ -0,0 +1,39 @@
...
+   * The discord server userId
+   */
+  userId: string | ObjectId;
+  /**
+   * List of previous "identity" information
+   * for a given user. Used to keep track of
+   * users incase they change their discord user name.
+   */
+  identities: Array<UserIdentity>;
+}
```
naomi-lgbt: So this will be a separate model from the user's suspension history?

src/db/moderator.ts
```diff
@@ -0,0 +1,49 @@
...
+ * actions taken by an admin.
+ */
+export interface Moderator {
+  /**
+   * The discord server userId for the mod
+   */
+  userId: string | Snowflake;
+  /**
+   * The discord server nickname of the mod
+   */
```
naomi-lgbt: On this one, I believe the `user` value is a Snowflake, but storing the `user.id` value is a string. We should type it as only one or the other to enforce consistency in the codebase, right?

src/db/user-suspension.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * The parent user
+   *
+   * **NOTE** I'm unsure if we should keep this as the objectId reference
+   * or use the discord provided snowflake
+   */
+  user: string | ObjectId | User;
+  /**
+   * The reason the user was suspended.
+   */
```
naomi-lgbt: Again, I think we need to determine what value will fill this field so we can be more strict with the typing.

naomi-lgbt (commented):
I ran these changes locally and because nothing has changed in the workflow yet I'm having trouble visualising how everything will come together. 😅 

I added some comments on a few things, but overall it looks good to me - I'm curious to see what will get saved when I suspend a user, and what I see when I call the `user` command.

src/db/moderator.ts
bradtaniguchi: Ah I didn't know that good call! I'll update the comments and schema to set this as an empty string as default.

src/db/user.ts
bradtaniguchi: So the idea I had was the `User` would be used to identify user's from Discord and possibly FCC itself. 

The suspension history/logging history/whatever information will be saved in other schemas by itself. This should make the database more flexible if we end up adding more stuff. Everything will have to have the same reference to the original user.

**I wasn't sure about the `userId` being different than the mongodb generated _id**, and was thinking if we should maybe make them 1 in the same?

src/db/moderator.ts
bradtaniguchi: I might of left this `Snowflake` in here on accident.

src/db/user.ts
naomi-lgbt: Discord ID values are theoretically unique, so that shouldn't be a problem. The only concern I would have there is future-proofing (if Discord runs out of numbers or switches how they identify a user, we'd have to change the `_id` value and I'm not sure mongo likes that).

bradtaniguchi:
Ok as a quick note, I looked into the twitter snowflake format (what discord users) more, and it appears to be a **string of numbers with the length of 18**. I'll be updating the `userId` property to the snowflake definition, **and change the `user` `_id`** to match the snowflake definition.

This way if we have the discord snowflake ID, we can lookup anything as-if we had the user `_id` as they will be the same.

bradtaniguchi:
So I updated the logic, and added 2 new "query" functions (name of these types of functions are up to debate). These functions essentially "do" stuff to the DB automatically for the 2 current "use-cases". 
- create a user-suspension
- create a user-log. (this command hasn't been created yet)

Both of these functions aren't integrated yet, nor is there any support to read this data, this probably will have to be another PR, to leverage what we have. 

**edit** anyone is welcome to read/review the code and provide feedback. I'll try to get into the integration of at least the user-suspension logic, so we save the data to the new schema. Also going to try to use gitpod with a test database installed there and see how the usability is :D

naomi-lgbt:
Wait this is the DB refactor - do we want to toss this for now?

bradtaniguchi:
This refactor we **can** merge, but its not hooked up to the codebase still. I think migrating the database later is more of a pain so getting this merged to master, even if its not functioning, would be better.

I would assume the "db features" wouldn't be supported until later tho

naomi-lgbt:
Sounds good - I'm good with the merge when you are :)

PR #158, opened 2020-08-26
Feature/db refactor

naomi-lgbt:
**Description:**

As discussed in the discord server, the suspend log embed currently uses the user mention. If the user leaves the server, or if the Discord client isn't able to fetch the data, the mention switches to `invalid-user` or the user ID string. 

I've updated the log embeds to include the user's username as well.

Closes #XXXXX


src/commands/suspend.ts
```diff
@@ -76,7 +76,7 @@ export const suspendCommand: CommandDef = {
         .addFields(
           {
             name: 'What happened?',
-            value: `${mod} has suspended ${user}.`
+            value: `${mod} has suspended **${user.user.username}** (${user}).`
```
bradtaniguchi, line 79: The fact its `user.user.username` is **weird**, our `user` might be mis-named, but we can handle this later.

naomi-lgbt: Yeh - this was an issue I ran into on my bot. 

The `user` variable is actually the `member` object, so the variable *should* be renamed as such.

bradtaniguchi:
@nhcarrigan Can you please look into fixing the commit history from your fork? It might require you to force update your history or even nuking your fork? (don't do this with open PRs tho)

Otherwise this will be fine, since we need this fix in "production" :D

**edit** I just realized, this should be targeted to master, not the PR

naomi-lgbt:
Yeh it's time for another nuke I think XD

This *should* be the only open PR from my fork so I can clean it up after this is merged :D

bradtaniguchi:
Damn, this PR was done off the wrong branch. Going to close since its going to get messy haha.

(already talked with nhcarrigan about opening a new PR)

PR #165, opened 2020-08-27
Update ESLint Rules

### LINKED ISSUE #144: [CHORE] Update lint rules to prevent single line ifs
**What needs to be done**
Update the eslint settings to prevent single line if statements.

**Additional context**
This will increase the verbosity of the code, but will prevent any annoying issues with linting and errors in the future. While doing #119 I had to change a number of `if` statements due to how the TS types are setup for most command, where a single line if wasn't possible to use.

naomi-lgbt:
**Description:**

This PR adds the `curly` rule, which requires that code blocks always use braces (no more single-line ifs). Had to also clean up the `suspend.ts` file from the master branch, which may cause issues with the merge of the DB refactor. 

@bradtaniguchi if you'd prefer, I can switch this to target the DB refactor branch and apply the rule after that gets merged.

Closes #144

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,6 +27,7 @@ module.exports = {
         ignoreReadBeforeAssign: false
       }
     ],
-    eqeqeq: ['error', 'always']
+    eqeqeq: ['error', 'always'],
+    curly: ['error']
   }
 };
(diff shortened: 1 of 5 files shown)


bradtaniguchi (approved):
LGTM :+1: :shipit: :rocket:

PR #145, opened 2020-08-16
Feat/119/update logger

bradtaniguchi:
**Description:**
Refactored all `console` calls to a winston logger.

Closes #119 

**note** this is a draft because the logging format isn't finalized yet, its currently just the default.

PR #170, opened 2020-08-30
Switch from Travis CI to GitHub Workflow

### LINKED ISSUE #160: [CHORE] Migrate from travis-ci to github-actions
**What needs to be done**
Currently we are using travis-ci to do a few things automatically. We should migrate to github actions instead.

**Additional context**
I'd like to do this mainly to keep everything in 1 place (github), and also we can leverage open source actions.

Please post here before opening any PR :)

Twaha-Rahman:
**Description:**

This PR adds the linting and testing part of the Travis CI checks in the GitHub workflows.

Closes #160


Twaha-Rahman:
@bradtaniguchi You'll have to set up Heroku deployment on your Heroku account and enable _Automatic Deploys_ from there.
The link below has all the details ->

https://devcenter.heroku.com/articles/github-integration#automatic-deploys

naomi-lgbt:
@bradtaniguchi - setting up automatic Heroku deployment is pretty seamless. When this is ready to merge, I can set up the deployment if you'd like. 

@Twaha-Rahman - Is it possible to include a `build` step in the new CI to ensure the code compiles before we merge to `master`?

Twaha-Rahman:
> @Twaha-Rahman - Is it possible to include a `build` step in the new CI to ensure the code compiles before we merge to `master`?

@nhcarrigan  Yes it is easily possible. But you'll not be able to transfer those built files to the master branch. Would you like me to add it?

naomi-lgbt:
Yeh there's no need to transfer the files - Heroku runs its own build. 

I think it would just be nice to have that extra "catch" step before we merge a PR, in case a contributor hasn't verified the `tsc` succeeds. 🙂

Twaha-Rahman:
@nhcarrigan Looks like I don't need add anything as Brad configured to run `tsc` after installing the dependencies. So the TS transpilation is also checked in the CI currently.

bradtaniguchi:
I'm unsure about splitting up the lint and test jobs. They could just be done one after another in-order. Seems like most stuff will be duplicated without much gain. Any thoughts? 

**edit** while testing out changes in the PR #145, with commit 481dc11d787e2a36fcf7da4d5d95b0f76669cd37 I realized checking travis-ci **and** LGTM is kinda annoying. It might be worth looking into LGTM integration to a github-action. This way we will only need to go to 1 place for tests rather than multiple. This doesn't have to be integrated with this PR, but could be an optimization made later :)

Twaha-Rahman:
> I'm unsure about splitting up the lint and test jobs. They could just be done one after another in-order. Seems like most stuff will be duplicated without much gain. Any thoughts?

@bradtaniguchi I split them up to two separate actions so that it is easy to understand what fails. If there's a linting error, then only the lint check will fail. If the tests failing, then only the test check will fail.

Previously in the Travis setup, if linting failed, then the tests won't run. But I think we should run both linting and testing even if the other one fails. That's why I split them up. What's your view?

bradtaniguchi:
> @bradtaniguchi I split them up to two separate actions so that it is easy to understand what fails. If there's a linting error, then only the lint check will fail. If the tests failing, then only the test check will fail.
> 
> Previously in the Travis setup, if linting failed, then the tests won't run. But I think we should run both linting and testing even if the other one fails. That's why I split them up. What's your view?

So there are 2 more steps (beyond any extra, like integrating LGTM to the pipeline)
- Add sanity build/deployment build
- Add deploy to heroku

With this in mind, I **don't** want to deploy to heroku, or run the build if there are "less" errors. However, I'm unsure if we can run all three of these things (lint,test,build) and if **any** of them fail, then we wont deploy. Currently idk if this is possible if we are to split up each action, but I'm pretty sure there is a way, either sequentially, or in parallel if we put all of this into 1 action pipeline.

naomi-lgbt:
We can leverage the GitHub-Heroku integration to allow automatic deploys from the master branch specifically. 

With this workflow, we can verify a PR passes all tests, merge it into master, and THEN heroku will fetch it and run the build and deployment.

Twaha-Rahman:
I was assuming make this workflow to just check the PRs and deployment can be made from the master branch on push like @nhcarrigan said.

@bradtaniguchi So would you like me to bring all tests under one workflow or leave it as is?

bradtaniguchi:
@nhcarrigan I'm reading about the github+heroku integration [here](https://devcenter.heroku.com/articles/github-integration#review-apps). As long as we can set it up to deploy on with commits made to master, then we should be back to travis-ci levels :D

@twaha-rahman its really a question of how much we want to maintain or how to deal with problems when we run stuff sequentially. Since I'd want us to move to adding more github actions in the near future, we either continue adding more workflows that all looks almost exactly alike and deal with multiple checks that run in parallel, or deal with 1 larger workflow that **could** only tell you part of the problem.

I'm currently leaning toward having a single workflow rather than multiple for simplicity sake. Yea we might have to deal with annoying issues thru multiple runs of CI checks, but **the checks are more a last line, not the first line**. The first line is running checks locally. Only LGTM might only be able to be ran in a github-action/check here, which I personally would want to change if possible.

naomi-lgbt:
image

As an example, my personal bot is configured to run automatic deploys from `master`. I work off a branch for new features, open the PR, let my CIs run, and after everything passes I merge. 

Then heroku deploys. The branch and PR don't get seen by heroku. :D

naomi-lgbt (approved):
LGTM!

bradtaniguchi (approved):
LGTM :+1:

Issue #168, opened 2020-08-30
[FEATURE] Remove automatic assignees to issue templates, keep them in PR


bradtaniguchi:
**What needs to be done**
Currently @bradtaniguchi and @nhcarrigan are automatically assigned as Assignees to most it not all issues created. This should be changed, to incentivies **other** user's to contribute to the bot. 😄  This will also fit with the workflow that stuff people work on is assigned to those working on it at that time.

**Additional context**
However the PR template should **keep** these two people as assignees, as they should review any PR's made.

PS. This issue is written in the third person, which is kinda funny haha

naomi-lgbt:
The PR template doesn't actually assign anyone. That's handled by a `CODEOWNERS` file which we don't have - easy enough to add though. 

LHLloyd has been eager to contribute to the bot and this seems like a good place to start. I'll ping him when he's next online and see if he's interested.

LandonLloyd:
I'll look into how to fix this! Thanks guys!!!

PR #171, opened 2020-08-30
Asignee fix

LandonLloyd:
**Description:**
I deleted the automatic assignees so that @bradtaniguchi and @nhcarrigan don't automatically get assigned to every issue. I also created a CODEOWNERS file with @bradtaniguchi and @nhcarrigan as owners so they can assign and unassign people from issues.

This fixes the Remove automatic assignees to issue templates, keep them in PR issue posted by @bradtaniguchi.
Closes #XXXXX

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: '[Bug]'
 labels: bug
-assignees: 'bradtaniguchi, nhcarrigan'
+assignees:
 ---
 
 **Describe the bug**
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,6 @@ name: Bug report
 about: Create a report to help us improve
 title: '[Bug]'
 labels: bug
-assignees:
 ---
 
 **Describe the bug**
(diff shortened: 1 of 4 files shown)


naomi-lgbt (approved):
LGTM! Nice work!

FYI - the `CODEOWNERS` file assigns us to *pull requests*, not issues. :)

naomi-lgbt:
Closes #168

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -3,7 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: '[Bug]'
 labels: bug
-assignees: 'bradtaniguchi, nhcarrigan'
+assignees:
```
bradtaniguchi: I think this property can be removed completely, rather than defined, but with no value. (same goes for the other 2 files)

bradtaniguchi (changes requested):
See above comment. 

I usually only set as a `comment` instead of a `request changes`, however since this is already approved, I want to make sure it doesn't accidentally get merged ;D

.github/ISSUE_TEMPLATE/bug_report.md
naomi-lgbt: Had to confirm this (GH automatically adds the property even if empty)  - it *does* seem to work without the property tho :D

LandonLloyd: Alright Brad! Should be good to go now! I removed the property as a whole on all three files!

bradtaniguchi: Awesome thanks!

bradtaniguchi (approved):
LGTM :+1:

PR #174, opened 2020-08-31
Fix/close logging

naomi-lgbt:
**Description:**

Adds the channel ID to the `close` embed, in case we need it. 

Adds the message ID to the `deleted message` embed as well. 

Closes #XXXXX


bradtaniguchi (approved):
LGTM 👍

Issue #101, opened 2020-07-17
[Bug] Issues with reaction based formatting if the message isn't a programming language


bradtaniguchi:
**Describe the bug**
Currently with the bot if a given message isn't recognized as a language, the bot explodes, and **removes all emojis on the given message**. This technically is 2 problems.

**To Reproduce**
1. Create a message that wont format to any given language, example: `%%%`
2. Add a random reaction and then use the "robot" reaction on said message
3. Observe all previous reactions disappear, and the bot will error out.

**Expected behavior**
Existing reactions should stay, **and** the bot should return the given message, just without any formatting.

**Additional context**
This seems to be partially fixed with a check in the `isSupportedByPrettier` function to make sure the `languageName` is defined. As the error occurs due to that param being undefined.

The other issue is not all emojis should be removed.

cjcanlas01:
Regarding the issue that not all emojis should be removed. I think modifying this part:

https://github.com/bradtaniguchi/discord-bot-test/blob/d7c20b3b1244e70c04ec06a79846b6c4e8cca8d5/src/commands/bootstrap.ts#L43

To something like this should get the job done:

https://github.com/bradtaniguchi/discord-bot-test/blob/d7c20b3b1244e70c04ec06a79846b6c4e8cca8d5/src/commands/bootstrap.ts#L79

Issue #100, opened 2020-07-16
Remove express dependency


bradtaniguchi:
**Is your feature request related to a problem? Please describe.**
Currently the bot integrates a super small express app, which was added to support waking up the bot in heroku. This is no longer needed.

**Describe the solution you'd like**
Remove the express code, and dependencies.

**Additional context**
We can always add the express server back if needed.

JoshQuaintance:
I'll take care of this

Issue #99, opened 2020-07-16
Update issue templates


bradtaniguchi:
Currently we have 4 issue templates
- bug report
- feature or enhancement
- feature request
- needs further testing

we should simplify this to 2:
- bug report (keep the same)
- feature or enhancement request

We haven't really used the "needs further testing" template, and the feature and feature or enhancement requests are essentially exactly the same.

I thought about adding a chore template, but I think we don't need that right now.

PR #178, opened 2020-09-04
Automatically add issue and PR embeds

Twaha-Rahman:
**Description:**

This PR adds the auto embedding for PRs and issues in the bot. This feature was inspired by the Gitter chatroom.

The bot will now add embeds for any PRs and issues when mentioned in this format -> #344

(Please note the that if you're quote replying someone and _their_ message includes an issue/PR, no embed will be sent.)


naomi-lgbt (commented):
A couple of thoughts I had:
1. Should this be listening just for a `#number`? Or should we preface it with `GH #number` - otherwise we'll get embeds when someone asks "How do I pass user story #3?".
2. Can the embed be tweaked to identify whether the result is an `issue` or a `pull request`?
3. Depending on what we do with number 1... if we make this require a little more intentional call, should the bot delete my message when it generates the embed?

Twaha-Rahman:
@nhcarrigan Here's my view -

> 1. Should this be listening just for a `#number`? Or should we preface it with `GH #number` - otherwise we'll get embeds when someone asks "How do I pass user story #3?".

I was thinking of limiting this functionality of the bot to only the `#contributor` channel. I also intended the feature to be passive feature rather than a command.... I did so because I think it would be a good feature to have the bot link embeds to that issue or PR when contributors are talking about it, rather than forcing them to use command. It's kinda like an auto-linking feature.

> 2. Can the embed be tweaked to identify whether the result is an `issue` or a `pull request`?

That can be done quite easily.

> 3. Depending on what we do with number 1... if we make this require a little more intentional call, should the bot delete my message when it generates the embed?

No. 1 question has the answer for this one.

naomi-lgbt:
My concern with the auto-linking is the risk of false positives. I wasn't thinking of making it a full command, but rather have it check for `gh #` instead of just `#`.

Twaha-Rahman:
> My concern with the auto-linking is the risk of false positives. I wasn't thinking of making it a full command, but rather have it check for `gh #` instead of just `#`.

I don't think there's much chance of a false positive if we limit the feature to only the `#contributor` channel. But if you really want to be safe, I could make the embedder ignore the issues from 1 - 50. That way nobody can accidentally trigger it. What do you say?

Twaha-Rahman:
Here's the updated preview -> https://discordapp.com/channels/705679411936100416/710976322721284167/751529788279226388

Twaha-Rahman:
Now the embedder ignores all issue/PRs that have numbers lower than 50. So no chance of false positive.

bradtaniguchi:
Back in gitter any time you typed `#` it was **auto-linked** just like this. I personally didn't like the feature, since I never wanted to reference an issue. 

If the feature was limited to a single channel (contributors) this might solve the "false positive" problem, however it turns more into a "magical feature" in that this feature is available to 1 and only 1 channel, which isn't very clear. 

The prefix approach might work if its **just** `gh #num`, that way its vastly clearer to use and only requiring 3 extra key strokes, then it could be available in any channel. 

However supporting another prefix/set of commands starts leading to another question, shouldn't we want to support **other** github types commands for other stuff and how would that look like? Would that even make any sense to add in the future too? 🤔

Twaha-Rahman:
> Back in gitter any time you typed `#` it was **auto-linked** just like this. I personally didn't like the feature, since I never wanted to reference an issue.

I was going for a feature like that. The reason being, remembering the bot command for linking issues is a mental overhead. I don't see contributors using `gh #num` syntax because they might not always remember it. Furthermore, in the long run when we'll move from Gitter to Discord all new contributors will come to Discord to get help and I don't see them knowing the bot command for linking an issue.

The intention behind the feature was to link issues/PRs behind the scenes by the bot **without** any extra efforts from the contributors themselves. Contributors should be chat normally and bot will automatically link those issues.

Also, I've tested this in Gitter. The same auto linking feature is used in Gitter and daily contributors are already used to using it. I don't want to introduce the same feature their used to with a different syntax, that's all. :sweat_smile: 

> If the feature was limited to a single channel (contributors) this might solve the "false positive" problem, however it turns more into a "magical feature" in that this feature is available to 1 and only 1 channel, which isn't very clear.

This problem was solved by ignoring any issue number that is lower than 50. fCC's currently open PR/issues numbers are currently way higher than 50. I don't think any contributor will be referencing an issue with a number that's under 50.

> The prefix approach might work if its **just** `gh #num`, that way its vastly clearer to use and only requiring 3 extra key strokes, then it could be available in any channel.

As I've already said, I think having the bot auto link issues and PRs was the goal. Contributors shouldn't need to do any extra stuff to get a basic feature like this.

> However supporting another prefix/set of commands starts leading to another question, shouldn't we want to support **other** github types commands for other stuff and how would that look like? Would that even make any sense to add in the future too?

@bradtaniguchi Yes, I'm planning to add more GitHub related features (after the migration ofc).

naomi-lgbt:
image

I saw your test in Gitter (which helps, because I'm 100% unfamiliar with gitter). 

The Discord API won't allow the bot to edit your message and insert a hyperlink, AFAIK - but is that all the gitter integration does? No embed or anything? Just a link? (A kind of broken link, tbh)

That's a *lot* less intrusive than the full-fledged embed we're using here. So it makes sense for *gitter*, because it's a silent feature. But for Discord, will it create too much noise? Especially if the bot is throwing an embed for things like "And reason #99 why I love Angular". Gitter would just turn my `#99` into a link - but our bot is sending an actual message, which is why I think it should require some *intent*. I should not accidentally get an embed from the bot. Our users should not accidentally get an embed from the bot. It comes off just a bit too spammy, IMO.

Twaha-Rahman:
@nhcarrigan How about restricting auto linking feature to only the #contribitor channel and requiring `gh #22` syntax on all other channels? I think this should help cut down on the false positives. Plus, I'll make the auto linking less intrusive by sending a small message like #222 that'll be linked to that issue (bots can use that markdown syntax even though users can't). And I think manual linking should show the full embed.

Any thought?

naomi-lgbt:
Wait, bots can use the hyperlink feature outside of an embed? 😮 

But yeh, that sounds like a better approach to me.

Twaha-Rahman:
@nhcarrigan Ok. So I'll implement it and let you guys know when I'm done.

bradtaniguchi:
Just a note, I've been playing around with these changes to see if we can throw out the current manual types, as I'm **sure** there are first party types somewhere out there. (Currently looking at `@octokit/types`)

---

Sounds like there will be 2 features in this PR.
- auto-linking on a single issue reference IE: `#99` will become a hyperlink to issue `#99`
- "explicit"-linking on a single issue with something like `gh #99` which will embed more information.

I was thinking about `gh` and how its totally different than other commands and realized we could also just integrate this into current `fCC` commands like:
```
fCC issue #99
```

Obviously this is vastly longer, however it will allow us to support much more stuff in the future, the most obvious of which is `pr` support. (I don't think the octokit knows if what is given is an issue or PR without making extra checks)
So we could end up supporting commands like:
```
fCC pr #100
```

Regardless of the approach taken for the 2nd feature, I think **this** PR should focus on **just the linking without the embed**. The embeds logic can **stay** in this PR, but it shouldn't be used. If you want to rip it out and open another PR for the 2nd part of the feature be my guest.


**Finally, please open issues to discuss features before open PR's in the future**. Unless its an obvious quick fix, or a small type/bug-fix odds are some discussion needs to be made, thus bloating a current PR's history and changes.

src/utilities/issue-embed-generator.ts
```diff
@@ -0,0 +1,36 @@
...
+  const {
+    title,
+    user,
+    comments,
+    state,
+    html_url,
+    pull_request,
+    number
+  } = issueData;
+  const bannerImageLink =
```
bradtaniguchi, line 17: Is this the best/most stable way to get this file? This seems like one of those "magic hard coded strings" thru github user content 🤔

src/utilities/get-issue-data.ts
```diff
@@ -0,0 +1,19 @@
+import { Octokit } from '@octokit/rest';
+import { IssueDef } from '../APIs/issue-def';
+
+export function getIssueData(issueNumber: number): Promise<IssueDef> {
```
bradtaniguchi: So I figured out that using the [@octokit/type](https://www.npmjs.com/package/@octokit/types) provides all the types we need, however using it is **interesting**. I can see why its in another library.

This is this full fill I created that removed the current manual `IssueDef` and provides first part type support for current and future endpoints.

I also refactored to code to rely more on inferred types with full TS coverage:
```typescript
import { Octokit } from '@octokit/rest';
import { Endpoints } from '@octokit/types';

// This can be refactored to what we needed, or made more specific,
// or renamed to be more clear? This also can be moved to a common file.
export type IssueEndpoint = Endpoints['GET /repos/:owner/:repo/issues/:issue_number'];

export const getIssueData = (
  issueNumber: number
): Promise<IssueEndpoint['response']> =>
  new Octokit().request('GET /repos/{owner}/{repo}/issues/{issue_number}', {
    owner: 'freeCodeCamp',
    repo: 'freeCodeCamp',
    issue_number: issueNumber
  });
```

bradtaniguchi (commented):
I figured out how to integrate first party types, we should migrate to that rather than the manual definitions.

Twaha-Rahman:
> I was thinking about `gh` and how its totally different than other commands and realized we could also just integrate this into current `fCC` commands like:
> 
> ```
> fCC issue #99
> ```
> 
> Obviously this is vastly longer, however it will allow us to support much more stuff in the future, the most obvious of which is `pr` support. (I don't think the octokit knows if what is given is an issue or PR without making extra checks)
> So we could end up supporting commands like:
> 
> ```
> fCC pr #100
> ```

The current embedder already can already distinguish between PRs and issues. According to GitHub API docs you'll have to check if the response has the `pull_request` key, if it does have it, then it's a PR. Otherwise, it's an issue.

> **Finally, please open issues to discuss features before open PR's in the future**. Unless its an obvious quick fix, or a small type/bug-fix odds are some discussion needs to be made, thus bloating a current PR's history and changes.

@bradtaniguchi Sorry for the trouble... I thought you guys were in a hurry with the migration and that's why I kinda threw everything together as fast as I could and completely overlooked the problems. 😅

Twaha-Rahman:
@bradtaniguchi Here's the new less intrusive auto linking feature -> https://discordapp.com/channels/705679411936100416/710976322721284167/752181666759573546

(I later on found that hyper link support for bots was removed recently.... so this was the only solution)

src/utilities/issue-embed-generator.ts
```diff
@@ -0,0 +1,48 @@
...
+      .setURL(html_url)
+      .setImage(bannerImageLink)
+      .setFooter(
+        "Thank you for your helping hand in freeCodeCamp's development!"
+      );
+  } else {
+    issueEmbed = new MessageEmbed()
+      .setTitle(`${embedType} #${number}`)
+      .setColor(state === 'closed' ? 'RED' : 'GREEN');
+  }
```
naomi-lgbt: This small embed doesn't include the link to the issue/PR. That could be added with:
```suggestion
  } else {
    issueEmbed = new MessageEmbed()
      .setTitle(`${embedType} #${number}`)
      .setColor(state === 'closed' ? 'RED' : 'GREEN')
      .setURL(html_url);
  }
```
Right?

src/utilities/issue-embed-generator.ts
naomi-lgbt: I'm not seeing where it's sent down in the API, tho. 🤔

src/utilities/get-issue-numbers.ts
```diff
@@ -0,0 +1,23 @@
...
+    issueStrings.forEach((issue) => {
+      const issueNumberOrNaN = Number(issue.substr(1));
+
+      if (!isNaN(issueNumberOrNaN)) {
+        if (issueNumberOrNaN > 50) {
+          issueNumbers.push(issueNumberOrNaN);
+        }
+      }
+    });
+  }
```
naomi-lgbt: So I peeked at the tests you were running - this code block means if I do `#100 #101 #102` the bot will send three links?

src/utilities/issue-embed-generator.ts
Twaha-Rahman: Oh sorry. Forgot about it 😓

src/utilities/get-issue-numbers.ts
Twaha-Rahman: Yes. But I don't think anybody will do that tbh.

Twaha-Rahman:
> I'm not seeing where it's sent down in the API, tho. :thinking: 

@nhcarrigan  On line 19 on the `issue-embedder.ts` file.

naomi-lgbt:
> > I'm not seeing where it's sent down in the API, tho. 🤔
> 
> @nhcarrigan On line 19 on the `issue-embedder.ts` file.

Sorry - I meant where does the GitHub API send *us* the image url? I think Brad was concerned about hard-coding the image url into the file, but I don't see a value when I call the API so I'm not sure how we'd get the image *without* hard-coding it?

Twaha-Rahman:
> Sorry - I meant where does the GitHub API send _us_ the image url? I think Brad was concerned about hard-coding the image url into the file, but I don't see a value when I call the API so I'm not sure how we'd get the image _without_ hard-coding it?

@nhcarrigan I just hard coded in that at that time.... I'm sure we'll be able to use some of fCC's CDN.

src/utilities/get-issue-data.ts
Twaha-Rahman: Added types from the package. Thanks Brad!

src/listeners/issue-embedder.ts
```diff
@@ -12,7 +12,7 @@ export function issueEmbedder(message: Message): void {
   if (issues.length > 0) {
     issues.forEach(async (issueNumber) => {
       try {
-        const issueData: IssueDef = await getIssueData(issueNumber);
+        const issueData: GithubResponse = await getIssueData(issueNumber);
```
bradtaniguchi: We don't need to explicitly provide the type here since its already provided from the return function. (Just realized this should be a lint rule 🤔  )

src/APIs/github-def.ts
```diff
@@ -0,0 +1,7 @@
+import { Endpoints } from '@octokit/types';
+
+type GithubEndpoints = Endpoints['GET /repos/:owner/:repo/issues/:issue_number'];
```
bradtaniguchi: This should be named more explicitly incase we end up getting more information. Maybe `GHIssueEndpoint`?

src/listeners/issue-embedder.ts
```diff
@@ -0,0 +1,25 @@
...
+import { issueEmbedGenerator } from '../utilities/issue-embed-generator';
+import { GithubResponse } from '../APIs/github-def';
+import { logger } from '../utilities/logger';
+import { getIssueData } from '../utilities/get-issue-data';
+
+export function issueEmbedder(message: Message): void {
+  const issues = getIssueNumbers(message.content);
+
+  if (issues.length > 0) {
+    issues.forEach(async (issueNumber) => {
```
bradtaniguchi: This setup has the problem of potentially returning issues out of order. Change to a `.map` + `Promise.all` so we can keep track of the order. 

Also I think maybe we should put a limit on how many can returned at the same time? Otherwise it would be easy to spam. (Maybe 3-5, or set as an environment variable)

src/utilities/get-issue-numbers.ts
bradtaniguchi: This can be changed to just:
```ts
if (!issueStrings) {
  // no matches
  return [];
}
return issueStrings
  // convert all strings to numbers, or NaN
  .map((issueString) => Number(issueString.substr(1)))
  // remove any non-numbers
  .filter((issueNum) => !isNaN(issueNum));
```

What cases could an `issueString` **not** be a number though? The regex should catch it right?

Also, I don't remember what the reasoning behind checking against `50` was.

src/utilities/issue-embed-generator.ts
```diff
@@ -0,0 +1,49 @@
...
+      );
+  } else {
+    issueEmbed = new MessageEmbed()
+      .setTitle(`${embedType} #${number}`)
+      .setColor(state === 'closed' ? 'RED' : 'GREEN')
+      .setURL(html_url);
+  }
+
+  return issueEmbed;
+}
```
bradtaniguchi: Remove all the `let` variables and use ternaries and early returns

src/utilities/get-issue-numbers.ts
naomi-lgbt: Checking against 50 was to avoid accidental misfires when I say something like "User Story #12 is asking you to centre your image"

src/listeners/issue-embedder.ts
naomi-lgbt: Personally I don't see a need to link more than one issue at a time.

src/utilities/get-issue-numbers.ts
bradtaniguchi: hmmm seems like another "magical feature/implicit feature"

Twaha-Rahman: This `if` statement can be removed if the bot is only auto linking in the `#contributors` channel. What do you guys say?

Twaha-Rahman:
> What cases could an `issueString` not be a number though? The regex should catch it right?

@bradtaniguchi There was only one case where the string would not be a number in the previous Regex. If it was a string like this -> "#100DaysOfCode" it would only match "#100" and load that issue.

I've updated the Regex now to account for such patterns. You can test it for yourself. Here's the new Regex -

```js
#[\d]+\s
```

### **UPDATE:**

That Regex wasn't matching any issue that was _at the end of a sentence_. I've updated the regex now and this fixes all the probs.

```js
#[\d]+(\s|$)
```

src/listeners/issue-embedder.ts
Twaha-Rahman: I've limited the auto linking to only 5 items by default and it's customizable by changing the env variables.

Twaha-Rahman:
> Personally I don't see a need to link more than one issue at a time.

@nhcarrigan I was hoping this would allow contributors to jump right into any particular issue/PR right away without the need for  searching it on GitHub. Let's say somebody says the following (or any sentence with more that one issue) ->

"I was working on #1888 and waiting for #3444."

In this case I would like to generate links for those issues. that's why I think limiting auto linking to only one issue per message wouldn't be a good idea.

src/utilities/get-issue-numbers.ts
bradtaniguchi: I'm on the side of **making this configurable** with a new environment variable for "issue_tag_channel", or similar. 

1. If the "issue_tag_channel" environment variable is passed, then the bot will listen for any issue numbers, including sub 50 numbers in that channel specified channel if it exists. 
2. If the "issue_tag_channel" environment variable is **not** passed, or the channel doesn't exist, then the bot will listen to all channels for any issue numbers.

This way we can easily tweak the behavior if we find the contributes only channel approach is to limiting, or if we find it more annoying. I think we can remove the "sub 50" limit since it seems to vague of a number to make much sense.

Twaha-Rahman:
@bradtaniguchi Umm... I did the required changes a few days ago. Forgot to tell you 😅

bradtaniguchi:
> @bradtaniguchi Umm... I did the required changes a few days ago. Forgot to tell you sweat_smile

Hey sorry, been lazy/busy. I'll try to check this out in the next day or so :)

src/listeners/issue-embedder.ts
```diff
@@ -0,0 +1,28 @@
...
+    try {
+      const { data } = await getIssueData(issueNumber);
+
+      const issueEmbed = issueEmbedGenerator(data, false);
+
+      message.channel.send(issueEmbed);
+    } catch (error) {
+      logger.error(error);
+    }
+  }
```
bradtaniguchi, line 29: This could end up putting the issues out of order depending on the `await` right? Please use `Promise.all` so they stay in the correct order.

src/listeners/issue-embedder.ts
```diff
@@ -0,0 +1,28 @@
...
+import { issueEmbedGenerator } from '../utilities/issue-embed-generator';
+import { logger } from '../utilities/logger';
+import { getIssueData } from '../utilities/get-issue-data';
+
+export async function issueEmbedder(
+  message: Message,
+  autoLinkLimit: number
+): Promise<void> {
+  const issues = getIssueNumbers(message.content);
+  if (issues.length === 0) return;
```
bradtaniguchi: Keep with the no single line lint.

.eslintrc.js
```diff
@@ -28,7 +28,6 @@ module.exports = {
       }
     ],
     'no-console': 'error',
-    eqeqeq: ['error', 'always'],
-    curly: ['error']
+    eqeqeq: ['error', 'always']
```
bradtaniguchi: Please don't remove lint rules.

src/listeners/issue-embedder.ts
Twaha-Rahman: It won't put issues out of order with this codeblock because it's not an iterative array method.

Here's link explaining it -> https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

.eslintrc.js
Twaha-Rahman: Ahh... sorry for removing the lint. I jt just felt odd writing a guard clause with curly brackets cause its normally written without them.

bradtaniguchi (approved):
If this looks how it should be, then it should be good to go! :+1: 
image

Twaha-Rahman:
Yup... that's how the auto linking feature will look like.

PR #179, opened 2020-09-04
Add message.send to error handling

### LINKED ISSUE #164: [CHORE] - Consistent Error Handling
**What needs to be done**

Currently we are inconsistent between logging error messages or sending a message to the end user. My proposed solution would be to ensure both occur during the error events.

For example, if a user calls the `suspend` command without being a moderator, the bot silently ignores them (and logs a message to the console). This makes the bot appear non-responsive or offline - I think we should ensure that the bot responds, even if it is as simple as "You are not authorised to use this command". 

**Additional context**

This could be handled as a separate PR, or tackled with #143 simultaneously.

naomi-lgbt:
**Description:**

I went through and hit all of our `if (x) logger.warn` steps to ensure the bot stills sends a response to the channel. This will prevent "silent errors", where the end user assumes the bot is broken because it does not reply.

Closes #164

### DIFF
--- a/src/commands/close.ts
+++ b/src/commands/close.ts
@@ -17,11 +17,17 @@ export const closeCommand: CommandDef = {
         logger.warn(
           `${message.author.username} did not have the correct permissions.`
         );
+        message.channel.send(
+          'Sorry, this command is restricted to moderators.'
+        );
         return;
       }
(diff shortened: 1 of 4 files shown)


bradtaniguchi (approved):
LGTM :+1:

PR #189, opened 2020-10-01
Add reaction-based role assignment

naomi-lgbt:
**Description:**

After Bjorno ran it by Quincy, the decision was made to add a role `Live Streams` to the channel. This role exists for users who wish to receive notifications when there is a live stream available on fCC (the YouTube, I think?). 

This PR adds a reaction feature to the bot that will allow us to post a message in `#rules`. Users can react with the 🎥 emoji and the bot will assign them the role and remove the reaction. Users who *have* the role can react again with 🎥 to have the role removed. 

The bot will need two new environment variables when this ships:
`STREAM_NOTIFY_ROLE` is the name of the role the bot should add/remove
`STREAM_MSG_ID` is the `id` of the message the users should react to (to prevent the event from firing on 🎥 reactions elsewhere). 

Closes #XXXXX

### DIFF
--- a/src/config/get-config.ts
+++ b/src/config/get-config.ts
@@ -81,6 +81,15 @@ export interface Config {
    * is set to `false`, then the feature will be available in all channels
    */
   AUTO_LINK_CHANNEL: string | false;
+  /**
+   * The role to assign to users who wish to be notified of new live streams.
+   */
+  STREAM_NOTIF_ROLE: string;
+  /**
+   * The message ID for the livestream role assignment - users react to this
+   * message to be assigned the role.
+   */
+  STREAM_MSG_ID: string;
 }
 /**
  * @name getConfig
@@ -103,6 +112,8 @@ export function getConfig(): Config {
     ONLINE_AT: getBotOnlineAt() || '',
     MOD_ROLE: process.env.MOD_ROLE || '',
     AUTO_LINK_LIMIT: Number(process.env.AUTO_LINK_LIMIT) || 5,
-    AUTO_LINK_CHANNEL: process.env.AUTO_LINK_CHANNEL || false
+    AUTO_LINK_CHANNEL: process.env.AUTO_LINK_CHANNEL || false,
+    STREAM_NOTIF_ROLE: process.env.STREAM_NOTIF_ROLE || '',
+    STREAM_MSG_ID: process.env.STREAM_MSG_ID || ''
   };
 }
--- a/src/config/get-config.ts
+++ b/src/config/get-config.ts
@@ -84,10 +84,11 @@ export interface Config {
   /**
    * The role to assign to users who wish to be notified of new live streams.
    */
-  STREAM_NOTIF_ROLE: string;
+  STREAM_NOTIFY_ROLE: string;
   /**
    * The message ID for the livestream role assignment - users react to this
-   * message to be assigned the role.
+   * message to be assigned the role. This ID is obtained by right clicking
+   * on the message you want to use for reactions and selecting "Copy ID".
    */
   STREAM_MSG_ID: string;
 }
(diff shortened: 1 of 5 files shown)


src/config/get-config.ts
```diff
@@ -81,6 +81,15 @@ export interface Config {
    * is set to `false`, then the feature will be available in all channels
    */
   AUTO_LINK_CHANNEL: string | false;
+  /**
+   * The role to assign to users who wish to be notified of new live streams.
+   */
+  STREAM_NOTIF_ROLE: string;
```
bradtaniguchi: Change to `STREAM_NOTIFY_ROLE`, not really worth it to save 1 character, or maybe its a typo :shrug: 

```suggestion
  STREAM_NOTIFY_ROLE: string;
```

src/config/get-config.ts
```diff
@@ -81,6 +81,15 @@ export interface Config {
...
+  /**
+   * The message ID for the livestream role assignment - users react to this
+   * message to be assigned the role.
+   */
+  STREAM_MSG_ID: string;
```
bradtaniguchi, line 93: How does one get this id again? Maybe we should update the comment rough instructions on how to get it.

src/config/get-config.ts
```diff
@@ -103,6 +112,8 @@ export function getConfig(): Config {
     ONLINE_AT: getBotOnlineAt() || '',
     MOD_ROLE: process.env.MOD_ROLE || '',
     AUTO_LINK_LIMIT: Number(process.env.AUTO_LINK_LIMIT) || 5,
-    AUTO_LINK_CHANNEL: process.env.AUTO_LINK_CHANNEL || false
+    AUTO_LINK_CHANNEL: process.env.AUTO_LINK_CHANNEL || false,
+    STREAM_NOTIF_ROLE: process.env.STREAM_NOTIF_ROLE || '',
```
bradtaniguchi:
```suggestion
    STREAM_NOTIFY_ROLE: process.env.STREAM_NOTIFY_ROLE || '',
```

src/reactions/livestream-reaction.ts
```diff
@@ -0,0 +1,34 @@
...
+      user = await user.fetch();
+    }
+    const target = reaction.message.guild?.member(user);
+    if (!target) {
+      logger.warn('user error');
+      return;
+    }
+    if (target.roles.cache.find((r) => r === streamRole)) {
+      target.roles.remove(streamRole).catch((e) => logger.error(e));
+      reaction.message.reactions.cache.get('🎥')?.remove();
```
bradtaniguchi, line 28: Is it part of the requirement to remove this instantly? Or should we actually keep it and maybe **remove** the user if they "un-react" to the message?

bradtaniguchi (changes requested):

src/config/get-config.ts
naomi-lgbt: That'll be something I slap in the .env once I post the message - after that we shouldn't have to touch it, but I can add instructions.

src/config/get-config.ts
naomi-lgbt: Haha `NOTIFY` is good - I was thinking of `NOTIFICAITONS` and didn't want to type that out

src/config/get-config.ts
naomi-lgbt: Updated the comment!

src/reactions/livestream-reaction.ts
naomi-lgbt: Do we want to do this refactor?

bradtaniguchi: as discussed we can add this later

bradtaniguchi (approved):
LGTM 👍

PR #191, opened 2020-10-02
Make thank feature togglable

naomi-lgbt:
**Description:**

Because the bot has access to the suspended channels, the brownie points feature triggers. This PR adds a config env to require the thanks feature be turned on before it will run. 

Closes #XXXXX


bradtaniguchi (approved):
LGTM :+1:

PR #193, opened 2020-10-04
Fix embed, use variable to clean up conditionals

### LINKED ISSUE #192: [Bug] Fix issue with invalid-user name appearing
**Describe the bug**
The same bug that affects other mentions also is affecting the "appeal" message sent to logs:

**Expected behavior**
We should put the current username within `()` or quotes or something after the id.

**Screenshots**
Screenshot 2020-10-04 at 11 54 00

**note** others see "invalid-user"

naomi-lgbt:
**Description:**

Changed the embed to use the `username` value instead of a direct mention. Also threw in a variable for `message.member.mentions?.first()`.

Closes #192

### DIFF
--- a/src/commands/close.ts
+++ b/src/commands/close.ts
@@ -52,10 +52,10 @@ export const closeCommand: CommandDef = {
         );
         return;
       }
-
+      const member = message.mentions.members?.first();
       let status = 'The appeal was not approved.';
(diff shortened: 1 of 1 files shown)


bradtaniguchi (approved):
LGTM :+1:

PR #194, opened 2020-10-14
Add Halloween feature

naomi-lgbt:
**Description:**

This PR adds a Halloween reaction feature, similar to the live stream feature. We will require two new `.env` values: `HALLOWEEN_ANNOUNCEMENT` and `HALLOWEEN_ROLE`. 

The PR itself should be fine to merge (as the env values are essential for the functionality), but the env values should not be added until we receive approval for the event.

Closes #XXXXX

### DIFF
--- a/src/config/get-config.ts
+++ b/src/config/get-config.ts
@@ -96,11 +96,13 @@ export interface Config {
    */
   THANK_OPTION: boolean;
   /**
-   * Announcement message ID for Halloween event
+   * Announcement message ID for Halloween event. This should be the ID
+   * of the message users can react to and receive HALLOWEEN_ROLE.
    */
   HALLOWEEN_ANNOUNCEMENT: string;
   /**
-   * Role for Halloween event.
+   * Role for Halloween event. The role grants access to the halloween event channel,
+   * which is the channel the Trick'cord Treat bot will be operating in.
    */
   HALLOWEEN_ROLE: string;
 }
--- a/src/config/get-config.ts
+++ b/src/config/get-config.ts
@@ -95,6 +95,14 @@ export interface Config {
    * Option to enable/disable the thanks feature.
    */
   THANK_OPTION: boolean;
+  /**
+   * Announcement message ID for Halloween event
+   */
+  HALLOWEEN_ANNOUNCEMENT: string;
+  /**
+   * Role for Halloween event.
+   */
+  HALLOWEEN_ROLE: string;
 }
 /**
  * @name getConfig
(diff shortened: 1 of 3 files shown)


src/reactions/halloween.ts
```diff
@@ -0,0 +1,34 @@
+import { logger } from '../utilities/logger';
+import { ReactionDef } from './reaction-def';
+
+export const halloweenReaction: ReactionDef = {
+  emoji: '🎃',
+  description: 'Adds the halloween role to the user who reacts',
```
bradtaniguchi: We should update this with more information for the command and event, like what does the Halloween role do?

src/reactions/halloween.ts
```diff
@@ -0,0 +1,34 @@
...
+  command: async (reaction, { config, user }) => {
+    const hallowRole = reaction.message.guild?.roles.cache.find(
```
bradtaniguchi, line 13: This whole action should be "skipped" if neither the `config.HALLOWEEN_ROLE` isn't defined, or the `config.HALLOWEEN_ANNOUNCEMENT` isn't defined.

src/reactions/halloween.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    if (reaction.message.id !== config.HALLOWEEN_ANNOUNCEMENT) {
+      return;
+    }
+    if (user.partial) {
+      user = await user.fetch();
+    }
+    const target = reaction.message.guild?.member(user);
+    if (!target) {
+      logger.warn('user error');
```
bradtaniguchi: can we add a more descriptive warning message?

src/config/get-config.ts
```diff
@@ -95,6 +95,14 @@ export interface Config {
    * Option to enable/disable the thanks feature.
    */
   THANK_OPTION: boolean;
+  /**
+   * Announcement message ID for Halloween event
+   */
+  HALLOWEEN_ANNOUNCEMENT: string;
```
bradtaniguchi, line 102: Update the comment with how is the announcement message used?

src/config/get-config.ts
```diff
@@ -95,6 +95,14 @@ export interface Config {
...
+  /**
+   * Role for Halloween event.
+   */
+  HALLOWEEN_ROLE: string;
```
bradtaniguchi, line 107: Update the comment with how the role is to be used

bradtaniguchi (changes requested):
Sorry for taking a lot longer than promised on reviewing this one!

src/reactions/halloween.ts
naomi-lgbt: Added additional description.

src/reactions/halloween.ts
naomi-lgbt: Added an initial check for those two variables, leveraging early return.

src/reactions/halloween.ts
naomi-lgbt: Filled it out a bit - this *shouldn't* actually trigger, tho. 
TS expects that `guild` might be undefined (which is the case for DMs) - including this check prevents compile errors in the later role assignment. 
But since it's locked to the specific message ID anyway, the command should never get this far in a DM.

src/config/get-config.ts
naomi-lgbt: Done.

src/config/get-config.ts
naomi-lgbt: Done.

bradtaniguchi (approved):
LGTM :+1:

PR #197, opened 2020-11-24
Create private command

naomi-lgbt:
Signed-off-by: nhcarrigan <nhcarrigan@gmail.com>

**Description:**

Creates a `private` command that generates a channel for moderators to speak with a specific user (without suspending them). 

Closes #XXXXX

### DIFF
--- a/src/commands/private.ts
+++ b/src/commands/private.ts
@@ -1,0 +1,132 @@
+import { logger } from '../utilities/logger';
+import { CommandDef } from './command-def';
+import { MessageEmbed, TextChannel } from 'discord.js';
+
+export const privateCmd: CommandDef = {
+  prefix: 'private',
+  description: 'Creates a private channel for moderators and the target user.',
+  usage: 'private <user>',
+  command: async (message, { config }) => {
+    try {
+      //check for appropriate permissions
+      if (!message.member?.hasPermission('KICK_MEMBERS')) {
+        logger.warn(
+          `${message.author.username} did not have the correct permissions.`
+        );
+        message.channel.send(
+          'Sorry, this command is restricted to moderators.'
+        );
+        return;
+      }
+
+      //check for log channel setting
+      const modChannel = message.guild?.channels.cache.find(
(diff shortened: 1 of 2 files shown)


src/commands/private.ts
```diff
@@ -0,0 +1,132 @@
...
+      //check for appropriate permissions
+      if (!message.member?.hasPermission('KICK_MEMBERS')) {
+        logger.warn(
+          `${message.author.username} did not have the correct permissions.`
+        );
+        message.channel.send(
+          'Sorry, this command is restricted to moderators.'
+        );
+        return;
+      }
```
bradtaniguchi: Is this the right permission check? Maybe we can use the `MOD_ROLE` environment variable check instead?

src/commands/private.ts
```diff
@@ -0,0 +1,132 @@
...
+      //logging embed
+      const restrictEmbed = new MessageEmbed()
+        .setColor('#FF0000')
+        .setTitle(`Private Channel Created`)
+        .addFields({
+          name: 'What happened?',
+          value: `${mod} has created a private discussion channel for **${user.user.username}** (${user}).`
+        })
+        .setFooter('Please remember to follow our rules!');
+      modChannel.send(restrictEmbed);
```
bradtaniguchi: Is this correct? Seems a lot like the normal suspension flow.

bradtaniguchi (commented):
Is this flow the same as the normal suspension flow, except we don't change the user's role? Might be worth thinking about refactoring these 2 workflows into 1

src/commands/private.ts
naomi-lgbt: That's a good approach, actually.

src/commands/private.ts
naomi-lgbt: Oh I should change the wording on the footer - but this is our mod-log embed and the action should still be logged I think 🤔

bradtaniguchi (approved):
LGTM :+1:

PR #199, opened 2020-11-25
feat/merge-close-suspend-private

naomi-lgbt:
Signed-off-by: nhcarrigan <nhcarrigan@gmail.com>

**Description:**

Bit of a big one here...
* Combines the `close`, `suspend`, and `private` command into one `moderate` command.
  * The three are now handled through a parameter for the new command
  * Command will not run at all if the user does not have a moderator role.
  * Command will not allow `suspend` or `private` to *target* a user with the moderator role.
* Renames the `mod` command to `modLinks` to avoid potential confusion.

Closes #XXXXX


src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+    if (!guild) {
+      logger.error('Guild error.');
+      return;
+    }
+
+    // Check for moderator role
+    const modRole = guild.roles.cache.find((role) => role.name === MOD_ROLE);
+    if (!modRole) {
+      logger.warn('Mod role not found.');
+      message.channel.send('Sorry, I could not find your moderator role.');
```
bradtaniguchi: Change to "Sorry, could not find moderator role". Since this isn't the user's moderator role, its a configuration issue. 

Unless we are following this way of addressing the chat? :thinking:

src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+    // Error handling
+    if (!guild) {
+      logger.error('Guild error.');
+      return;
+    }
+
+    // Check for moderator role
+    const modRole = guild.roles.cache.find((role) => role.name === MOD_ROLE);
+    if (!modRole) {
+      logger.warn('Mod role not found.');
```
bradtaniguchi, line 27: Like the next `if` statement, I think adding the user's username would help with debugging.

src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+      );
+      message.channel.send('Sorry, this command is restricted to moderators.');
+      return;
+    }
+
+    // Create arguments list, removing prefix and command name
+    const commandArguments = content.split(' ').slice(2);
+
+    // Get action
+    const action = commandArguments.shift();
```
bradtaniguchi: This might be done better using something like:
```js
const [action] = content.split(' ').slice(2);
```
(this assumes the length of the command is correct length tho)

src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+      return;
+    }
+
+    // Check for suspend role
+    const suspendRole = guild?.roles.cache.find(
+      (role) => role.name === SUSPEND_ROLE
+    );
+
+    if (!suspendRole) {
+      await channel.send('Sorry, but I could not find your suspended role.');
```
bradtaniguchi: add console.log, also why does this one have an `await`? :thinking:

src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+      return;
+    }
+
+    if (action === 'close') {
```
bradtaniguchi, line 64: This has to be moved to another function, otherwise this command gets way to large.

Pass in the "top-level" objects such as `config` and `message` to each sub-function and manage everything in there.

src/commands/moderate.ts
```diff
@@ -0,0 +1,305 @@
...
+      MOD_ROLE,
+      SUSPEND_CATEGORY,
+      BOT_ROLE
+    } = config;
+
+    // Error handling
+    if (!guild) {
+      logger.error('Guild error.');
+      return;
+    }
```
bradtaniguchi, line 22: How when is this possible? 🤔

bradtaniguchi (changes requested):
So I started to review but reviewing a 300 line function became to much. I suggest we refactor the command into 2 separate sub-functions, that are called based upon the `action`. This way the code should get cut down in half and it will make it easier to review and go through :)

also happy Thanksgiving 🦃

src/commands/moderate.ts
naomi-lgbt: Sure thing - I'll refactor.

src/commands/moderate.ts
naomi-lgbt: I recently learned that many things, including the `.send` call, in Discord are actually asynchronous.

src/commands/moderate.ts
naomi-lgbt: Maybe "could not find the server's moderator role"?

bradtaniguchi: Yea that should work out :+1:

src/commands/moderate.ts
naomi-lgbt: DM messages have no `guild` property, I believe.

This is an error handling trick I learned with my own bot, to avoid optional chaining everywhere that `guild` is accessed.

src/commands/moderate.ts
naomi-lgbt: This one doesn't pertain to the user that calls it though - it fires if the role name in the `.env` doesn't match a role in the server.

bradtaniguchi:
is this good to review again or still a WIP? No rush just checking :)

naomi-lgbt:
Good to review again! Sorry, should have said something 😅 

Rather than pass around the upper-level config, I created a separate interface for passing the actual channels/roles to the helper functions to avoid having to write the checks multiple times.

bradtaniguchi (commented):
Sorry for being super slow on this, I'll be willing to merge it but I wont be able to test it ATM

naomi-lgbt:
Actually we can hold off on the merge for now - it sounds like the suspension flow might need to be removed.

bradtaniguchi:
Ah ok :+1:

naomi-lgbt:
Okay this should be good for a review/merge whenever you have time :)

bradtaniguchi (approved):
Code looks good :+1:

So I guess now there is just a "private channel" flow right?

naomi-lgbt:
Correct - essentially the same as the former suspension flow, but without the role assignment. So this'll result in two commands:
```
!fCC moderate private @nhcarrigan
Creates a private chat with the tagged user and the moderator team

!fCC moderate close
Closes a `private-` channel.

PR #261, opened 2023-01-08
chore: Adding wrappers for guild and privileged commands

plamoni:
Consolidates logic for checking whether a command is executed within a guild and also checking for the required permissions for the command. This simplifies command implementation and reduces the opportunity for errors.

Checklist:

This PR is as much a proposal as anything. It has not yet been thoroughly tested (will perform more testing if the approach is approved).


bradtaniguchi (commented):
I'd add some comments to at least the high level types/functions added to help describe how to use it. Specifically the `wrapCommands` function, which has a name that explains what it does, but not why or what a "wrapped" command represents. 

Also what's the purpose of a "wrapped command", which is the default case when a command isn't a privileged or guild command?

plamoni:
> I'd add some comments to at least the high level types/functions added to help describe how to use it. Specifically the `wrapCommands` function, which has a name that explains what it does, but not why or what a "wrapped" command represents.

Will do.

> Also what's the purpose of a "wrapped command", which is the default case when a command isn't a privileged or guild command?

There's two categories of commands, "wrapped" and "unwrapped." Wrapped commands are commands that are safe to register, but unwrapped commands are commands that need additional work (what I call "wrapping") before they're safe.

For example, `PrivilegedCommand` and `GuildCommand` are unwrapped commands. They have metadata describing how they should be handled, but if you just registered one directly, then it wouldn't properly gate using that metadata (e.g. a `PrivilegedCommand` would not stop users from executing it if they lacked sufficient permissions).

`WrappedCommand` is only used to ensure that TypeScript prevents unwrapped commands from being registered (as the registration code expects only commands of that type).

Another approach I considered would be to use `Command` in place of `WrappedCommand` and then not have the unwrapped commands extend it. But the problem with that is that TS uses structural types and so unless the unwrapped command types are missing or do not match a required field from `Command`, then TS will actually accept them as `Command` objects, which makes them dangerous (since someone might accidently register them without wrapping them first). While both of the current unwrapped commands use a different type for `run` than `Command` does (and hence wouldn't be assignable to `Command` variables/collections), I don't really want to rely on that in the future. It feels more appropriate to create a clear distinction.

plamoni:
I've done some local testing to confirm this appears to work as expected. I haven't rigorously tested that each command has the right permissions (reviewer will just need to confirm that by looking at the code, as I have done). But I can confirm through local testing that the general gating approach does work:

image

I am not sure how to test issuing a command outside a guild. Would be happy to try it if someone can point me to how. :-)

src/utils/wrapCommands.ts
```diff
@@ -0,0 +1,89 @@
...
+): command is PrivilegedCommand => {
+  return (
+    isGuildCommand(command) &&
+    !!(command as PrivilegedCommand).requiredPermissions
+  );
+};
+
+const isGuildCommand = (command: UnwrappedCommand): command is GuildCommand => {
+  return !!(command as GuildCommand).guildOnly;
+};
```
naomi-lgbt:
```suggestion
const isPrivilegedCommand = (
  command: UnwrappedCommand
): command is PrivilegedCommand =>
  isGuildCommand(command) && "requiredPermissions" in command;

const isGuildCommand = (command: UnwrappedCommand): command is GuildCommand =>
  "guildOnly" in command;
```

Rather than coercing, we could check for the key `in` object?

src/utils/wrapCommands.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+export const wrapCommands = (
+  commands: UnwrappedCommand[]
+): WrappedCommand[] => {
+  return commands.map((command) => {
+    if (isPrivilegedCommand(command)) {
+      return wrapPrivilegedCommand(command);
+    } else if (isGuildCommand(command)) {
+      return wrapGuildCommand(command);
+    }
```
naomi-lgbt:
```suggestion
    if (isPrivilegedCommand(command)) {
      return wrapPrivilegedCommand(command);
    }
    if (isGuildCommand(command)) {
      return wrapGuildCommand(command);
    }
```

Nitpicking, but don't need `else if` on an early return.

naomi-lgbt (changes requested):
Some nitpicking

src/utils/wrapCommands.ts
plamoni: The reason I went with (and tend to lean towards) the use of `as` like this is because of [property renaming](https://developers.google.com/closure/compiler/docs/limitations#implications-of-global-variable,-function,-and-property-renaming:). While I realize we don't use  it for this project, I've come to consider it a best practice to write code that will not break when its enabled. By mixing string references and dot references to the property, your suggested approach would likely break when subjected to property renaming.

There's [no runtime performance advantage](https://jsbench.me/19lcurvvet/1) to either approach:

image

However, there _is_ a logical difference between the two approaches that's notable:

```js
const myObj = { prop: undefined };
console.log("prop" in myObj); // true
console.log(!!myObj.prop); // false
```

In the case of my code above, the type guard is doing double duty: checking both the presence _and_ truthiness of the target property. Due to how [`GuildCommand` is defined](https://github.com/plamoni/discord-bot/blob/5aa7b646f0562fce6c2c5aa327b62c7114f36964/src/interfaces/GuildCommand.ts#L6), an object with the `guildOnly` property set to anything other than `true` does not satisfy the `GuildCommand` interface. Therefore, a type guard checking only for the _presence_ of the property risks it mis-identifying a command with `guildOnly: false` as  being a `GuildCommand`.

To be fair, my solution does identify any truthy value in `guildOnly` as indicating satisfaction of the `GuildCommand` interface, though I think it's fair to say someone putting a truthy value in `guildOnly` reasonably intends it to be a guild-only command, while someone putting `false` in that field would be (with reason) surprised to find it was interpreted as a guild-only command.

Given that we've spent a combined several minutes on this one bit of code, I think it's only fair that I take _some_ action on it. Why not make it more resilient to runtime issues and more specifically accurate? e.g.:

```ts
const isGuildCommand = (command: UnwrappedCommand): command is GuildCommand =>
  (command as GuildCommand)?.guildOnly === true;
```

This has the advantage of being resilient to being passed a non-object argument (something that really could only happen by circumventing type checking) and also explicitly checking for a value of `true` instead of accepting any truthy value. It is also resilient to property renaming and omits the unnecessary block wrapper.

Let me know what you think and I'll make the change for both guards.

naomi-lgbt: Thanks for the detailed writeup - this makes quite a bit of sense.

I'd go ahead with the more resilient solution you proposed.

plamoni: Done.

naomi-lgbt:
Apologies for the delay - should have room to circle back and review again tomorrow.

plamoni:
> Apologies for the delay

No delay at all... I just got around to wrapping up the concerns this weekend. It's been like 1 work hour since I got the latest commits in. Review at your leisure. :-)

naomi-lgbt (approved):
Okay, I've given this some testing, and I think we're good to go. Thanks for the hard work on this one!

PR #274, opened 2023-02-10
chore: update private channel message

### LINKED ISSUE #273: Can we change "The moderation team have created this private channel..." message?
**Request**
Can we change the initial message when a private channel is created?

The current message is:
"The moderation team have created this private channel to discuss something with you."

Can we change it to something more general, for example:
"A private channel was created by [name]. This channel is visible to only you and the moderation team."

**Context**
I think the current message may seem scary at a first glance (seems like you have caused some moderation problems) so I've been limiting the use of PMs to moderation concerns.

However, I want to use PMs to communicate with contributors individually for more casual conversation. For example, helping new contributors who are afraid of asking questions in a public channel, or just asking if they need any help to increase engagement of the contributors.

I also think it's better to use our PM in my use case above, not Discord's DM. Discord's DM is outside of freeCodeCamp server and it can cause moderation problems.

sidemt:
Checklist:

Closes #273

I haven't tested this by actually running the bot. I thought the change is simple enough to just update the message, let me know if I'm missing anything.

### DIFF
--- a/src/commands/privateChannel.ts
+++ b/src/commands/privateChannel.ts
@@ -156,7 +156,7 @@ export const privateChannel: Command = {
       ]);
 
       await newChannel.send({
-        content: `Hey <@!${target.id}>!\n\nThe moderation team have created this private channel to discuss something with you.`,
+        content: `Hey <@!${target.id}>!\n\nA private channel was created. This channel is visible to only you and the moderation team.`,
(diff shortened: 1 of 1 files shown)


naomi-lgbt (approved):
Perfect, this should do the trick!

PR #324, opened 2023-04-11
Chore: Create a basic test workflow

### LINKED ISSUE #241: [chore] Add unit tests, or remove test from workflow?
Currently there is an `npm run test` call in the current workflow, but there are no tests. There is `jest` and `ts-jest` dependencies so maybe it might be time to look back into adding some basic unit tests? 

Or removing these tests could provide some cleanup.

a2937:
Checklist:

Closes #241

I created a starting test suite filled with a test for the millisecond calculation utility. I also decided to make a tiny adjustment to the calculate millisecond utility by making all units be converted to lower case and completing the typescript typing for it.


.eslintignore
naomi-lgbt, line 1: This seems to be from a different repo and not necessary here.

src/utils/calculateMilliseconds.ts
```diff
@@ -6,8 +6,8 @@
  * @param {string} unit The unit of time to convert to milliseconds.
  * @returns {number} The number of milliseconds.
  */
-export const calculateMilliseconds = (value: number, unit: string) => {
-  switch (unit) {
+export const calculateMilliseconds = (value: number, unit: string): number => {
+  switch (unit.toLowerCase()) {
```
naomi-lgbt, line 10:
```suggestion
export const calculateMilliseconds = (value: number, unit: "seconds" | "minutes" | "hours" | "days" | "weeks"): number => {
  switch (unit) {
```

We could actually get more strict with the typing and avoid potential issues.

package.json
```diff
@@ -30,6 +30,7 @@
   "devDependencies": {
     "@types/common-tags": "1.8.1",
     "@types/diff": "5.0.3",
+    "@types/jest": "^29.5.0",
```
naomi-lgbt:
```suggestion
    "@types/jest": "29.5.0",
```

Will also need to run `npm i` after committing this to update the package-lock.json

naomi-lgbt (changes requested):

.eslintignore
a2937: That's correct and completely my fault. I'll remove it.

src/utils/calculateMilliseconds.ts
a2937: Thank you.

package.json
a2937: Done.

a2937:
Looks like I need to move the test folder over

naomi-lgbt (changes requested):
We're almost there.

The `tsconfig.json` will need to have an `exclude` property added, set to an array that contains `./test`.

[Example](https://github.com/BeccaLyria/discord-bot/blob/main/tsconfig.json).

a2937:
:/ I just wanted to make a simple pull request.

src/utils/calculateMilliseconds.ts
a2937: However we can't actually do that as I just found out. Because the mute command will fail verification.

naomi-lgbt (approved):
Happy to land this as is. I'll circle back and re-implement the type def + write a type guard.

PR #349, opened 2023-05-12
feat: load testing library, run first tests!

naomi-lgbt:
Checklist:

- [ ] I have read [freeCodeCamp's contribution guidelines](https://contribute.freecodecamp.org).
- [ ] My pull request has a [descriptive title](https://contribute.freecodecamp.org/#/how-to-open-a-pull-request?id=prepare-a-good-pr-title) (**not** a vague title like `Update index.md`)

Closes #XXXXX

### DIFF
--- a/test/commands/ban.spec.ts
+++ b/test/commands/ban.spec.ts
@@ -1,21 +1,60 @@
 import { assert } from "chai";
 import { ApplicationCommandOptionType, PermissionFlagsBits } from "discord.js";
+import {
+  MockChannel,
+  MockChatInputCommandInteraction,
+  MockGuild,
+  MockMember,
+  MockUser,
+  MockWebhook,
+} from "discordjs-testing";
 
 import { ban } from "../../src/commands/ban";
+import { Camperbot } from "../../src/interfaces/Camperbot";
 
-suite("ban command", () => {
-  test("ban command is defined", () => {
+const guild = new MockGuild({ name: "freeCodeCamp" });
+const user = new MockUser({
+  username: "naomi",
+  bot: false,
+  discriminator: 1,
+  avatar: "https://cdn.nhcarrigan.com/profile.png",
+  system: false,
+});
+const bot = new MockUser({
+  username: "camperbot",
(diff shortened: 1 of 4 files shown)


test/commands/ban.spec.ts
```diff
@@ -1,21 +1,60 @@
...
+  avatar: "https://cdn.nhcarrigan.com/profile.png",
+  system: false,
+});
+const hook = new MockWebhook({
+  channel,
+  user: bot,
+});
+guild.members.add(target);
+
+suite.only("ban command", () => {
```
plamoni, line 45: Probably didn't mean to have `.only` on this when it's submitted.

test/commands/ban.spec.ts
```diff
@@ -39,4 +78,115 @@ suite("ban command", () => {
...
+    await ban.run({} as Camperbot, interaction.typeCast<"cached">());
+    assert.lengthOf(interaction.replies, 1);
+    assert.equal(interaction.replies[0].content, "I cannot ban them.");
+  });
+
+  test("should ban a member", async () => {
+    const targetMember = guild.members.cache.get(target.id);
+    if (!targetMember) {
+      assert.fail("Target member not found");
+      return;
```
plamoni, line 161: Is this necessary? I suspect if you do an `assert.fail` it will immediately terminate the test.

If you're adding the `return` just to allow TS to infer `targetMember` is set, then consider just using optional chaining or a null-safe assertion (e.g. `targetMember!.bannable = true`).

plamoni (commented):
Looks pretty good. I think there's some work that could be done to make the tests more readable. Also, mocking out MongoDB seems like something we'll need to do sooner or later. But overall a great start. Thanks!

plamoni (commented):
Looks pretty good. I think there's some work that could be done to make the tests more readable. Also, mocking out MongoDB seems like something we'll need to do sooner or later. But overall a great start. Thanks!

naomi-lgbt:
Closing to start again from a clean branch.

PR #565, opened 2024-03-20
fix : check existing labels in add label command

### LINKED ISSUE #531: [FIX] - The add label command should check existing labels
The command should only be able to add labels that are already present in the repository's settings.

a2937:
Checklist:

Closes #531

This issue has been created as a draft because I am not exactly sure where to get a GitHub token to properly test this command, but the logic seems fairly sound.

### DIFF
--- a/src/commands/subcommands/github/handleAddLabels.ts
+++ b/src/commands/subcommands/github/handleAddLabels.ts
@@ -18,38 +18,36 @@ export const handleAddLabels: Subcommand = {
       const labels = interaction.options.getString("labels", true);
       const labelNames = labels.split(",").map((l) => l.trim());
 
-      const response = await Bot.octokit.request(
-        "GET /repos/{owner}/{repo}/labels",
-        {
-          owner: "freeCodeCamp",
-          repo: repo
-        }
-      );
+      const response = await Bot.octokit.rest.issues.listLabelsForRepo({
+        owner: "freeCodeCamp",
+        repo
+      });
 
       const presentLabels = response.data.map((x) => x.name.trim());
 
       if (
-        labelNames.every((requestedLabel) =>
+        !labelNames.every((requestedLabel) =>
           presentLabels.includes(requestedLabel)
         )
       ) {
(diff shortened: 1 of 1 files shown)


src/commands/subcommands/github/handleAddLabels.ts
```diff
@@ -18,18 +18,38 @@ export const handleAddLabels: Subcommand = {
       const labels = interaction.options.getString("labels", true);
       const labelNames = labels.split(",").map((l) => l.trim());
 
-      await Bot.octokit.issues.addLabels({
-        owner: "freeCodeCamp",
-        issue_number: number,
-        repo,
-        labels: labelNames
-      });
+      const response = await Bot.octokit.request(
```
naomi-lgbt: We should be able to use `Bot.octokit.issues.listLabelsForRepo()` rather than constructing our own request.

src/commands/subcommands/github/handleAddLabels.ts
```diff
@@ -18,18 +18,38 @@ export const handleAddLabels: Subcommand = {
...
+          repo: repo
+        }
+      );
+
+      const presentLabels = response.data.map((x) => x.name.trim());
+
+      if (
+        labelNames.every((requestedLabel) =>
+          presentLabels.includes(requestedLabel)
+        )
```
naomi-lgbt: We should invert the logic. And leverage early return patterns.

if (label not valid) {
  interaction.editReply(label don't exist)
  return
}

add label
interaction.editReply(label added)

src/commands/subcommands/github/handleAddLabels.ts
a2937: I think I got it. The method should do exactly what the custom request did.

src/commands/subcommands/github/handleAddLabels.ts
a2937: Understood.

naomi-lgbt (approved):

PR #946, opened 2026-06-01
feat: let mods choose issue close reason (completed/not planned/duplicate)

majestic-owl448:
Checklist:

When closing an issue via the `/github close` command, CamperChan previously closed it without a `state_reason`, which caused GitHub to default to "completed" (shown as "resolved" in the UI). This PR lets moderators choose the close reason instead.

Two new optional options are added to `/github close`:

- `reason`: choose between `completed`, `not planned`, and `duplicate`. Defaults to `not planned`, so the existing behavior is preserved when the option is omitted.
- `duplicate-of`: the issue number this one duplicates. Required when `reason` is `duplicate`.

`completed` and `not planned` are applied via the REST `issues.update` endpoint. `duplicate` uses the GraphQL `closeIssue` mutation with `stateReason: DUPLICATE` and `duplicateIssueId`, which creates the real "marked as a duplicate of #N" linkage that REST cannot.

Validation (each replies in Discord and aborts before any change is made):

- Selecting `reason` or `duplicate-of` while closing a pull request is rejected, since these only apply to issues.
- Choosing `duplicate` without a `duplicate-of` number is rejected.
- A `duplicate-of` that points to a pull request is rejected, since the duplicate target must be an issue.

### DIFF
--- a/src/commands/subcommands/github/handleClose.ts
+++ b/src/commands/subcommands/github/handleClose.ts
@@ -58,10 +58,12 @@ export const handleClose: Subcommand = {
         owner:        "freeCodeCamp",
         repo:         repo,
         state:        "closed",
-        // eslint-disable-next-line @typescript-eslint/naming-convention -- Github API name.
-        state_reason: isPull
-          ? undefined
-          : "not_planned",
+        ...isPull
+          ? {}
+          : {
+            // eslint-disable-next-line @typescript-eslint/naming-convention -- Github API name.
+            state_reason: "not_planned" as const,
+          },
       });
       if (isPull && isSpam) {
         await camperChan.octokit.rest.issues.addLabels({
--- a/src/commands/subcommands/github/handleClose.ts
+++ b/src/commands/subcommands/github/handleClose.ts
@@ -62,7 +62,7 @@ export const handleClose: Subcommand = {
           ? {}
           : {
             // eslint-disable-next-line @typescript-eslint/naming-convention -- Github API name.
-            state_reason: "not_planned" as const,
+            state_reason: "not_planned",
           },
       });
       if (isPull && isSpam) {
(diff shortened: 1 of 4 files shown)


src/commands/subcommands/github/handleClose.ts
```diff
@@ -58,6 +58,12 @@ export const handleClose: Subcommand = {
         owner:        "freeCodeCamp",
         repo:         repo,
         state:        "closed",
+        ...isPull
+          ? {}
+          : {
+            // eslint-disable-next-line @typescript-eslint/naming-convention -- Github API name.
```
plamoni: Is this still needed given you updated the eslint config?

src/commands/subcommands/github/handleClose.ts
```diff
@@ -58,6 +58,12 @@ export const handleClose: Subcommand = {
...
+            state_reason: "not_planned" as const,
```
plamoni: nit: I'm a little surprised TS can't infer this string literal type based on context. Can you remove the `as const`? If you do and i causes an error, what's the error you get?

plamoni (commented):
Drive-by feedback from Pete... 😄 

Feel free to ignore me.

src/commands/subcommands/github/handleClose.ts
majestic-owl448: Removing the disable comment gives:

```
64:13  warning  Object Literal Property name `state_reason` must match one of the following formats: camelCase  @typescript-eslint/naming-convention
```

It fails linting.

a2937:
New Options available

Short of creating a mock repository to mess with; everything seems to be in order.

a2937:
This does make me wonder if this could be pretty useful. Something to consider perhaps. 

 https://github.com/octokit/fixtures/tree/main

majestic-owl448:
do I need to do any updates here?

PR #948, opened 2026-07-06
feat: prepopulate discord id in form

### LINKED ISSUE #947: Include user ID in ban message
### Is your feature request related to a problem? Please describe.

I've noticed that users frequently struggle to figure out their user ID when appealing a ban. This adds additional effort for mods who want to unban because we sometimes need to go hunting down the ID (it's frequently located in places that are hard to get to and copy from on mobile it seems).

### Describe the solution you'd like

I think we should update the [message sent to the user](https://github.com/freeCodeCamp/CamperChan-discord/blob/bcb8eb1e76f3ec512fce15aea5f0779256aa437c/src/modules/sendModerationDm.ts#L40) to include their user ID so they can just copy/paste it into the appeal form.

I believe this ID is available from `user.id`, which is in-context already on that handler.

### Describe alternatives you've considered

It would be extra-neat if we could pipe the ID all the way around somehow so that users don't even need to copy/paste. This [might be possible with Google Forms](https://support.google.com/docs/answer/2839588?hl=en&sjid=554345920199269984-NC#zippy=%2Csend-a-form-with-pre-filled-answers). Though I haven't tested it out to be sure.

### Additional context

_No response_

a2937:
Checklist:

Closes #947

### DIFF
--- a/src/modules/sendModerationDm.ts
+++ b/src/modules/sendModerationDm.ts
@@ -37,7 +37,7 @@ export const sendModerationDm = async(
     if (action === "ban") {
       embed.addFields({
         name:  "Appeals",
-        value: `You can use [this form](https://forms.nhcarrigan.com/o/docs/forms/k7fPZmhgCS4RRKbStq8AcW/4?Discord_ID=${user.id}&?Username=${user.username}) to appeal your ban after you have [read our code of conduct](https://www.freecodecamp.org/news/code-of-conduct).`,
+        value: `You can use [this form](https://forms.nhcarrigan.com/o/docs/forms/k7fPZmhgCS4RRKbStq8AcW/4?Discord_ID=${user.id}&Username=${user.username}) to appeal your ban after you have [read our code of conduct](https://www.freecodecamp.org/news/code-of-conduct).`,
       });
     }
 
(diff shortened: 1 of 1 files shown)


majestic-owl448 (commented):
it's not being filled in, Naomi will need to turn this setting on

image


let's just include it in the DM that CamperChan is sending out for now

src/modules/sendModerationDm.ts
```diff
@@ -37,7 +37,7 @@ export const sendModerationDm = async(
     if (action === "ban") {
       embed.addFields({
         name:  "Appeals",
-        value: `You can use [this form](https://forms.nhcarrigan.com/o/docs/forms/k7fPZmhgCS4RRKbStq8AcW/4?Discord_ID=${user.id}) to appeal your ban after you have [read our code of conduct](https://www.freecodecamp.org/news/code-of-conduct).`,
+        value: `You can use [this form](https://forms.nhcarrigan.com/o/docs/forms/k7fPZmhgCS4RRKbStq8AcW/4?Discord_ID=${user.id}&?Username=${user.username}) to appeal your ban after you have [read our code of conduct](https://www.freecodecamp.org/news/code-of-conduct).`,
```
majestic-owl448:
```suggestion
        value: `You can use [this form](https://forms.nhcarrigan.com/o/docs/forms/k7fPZmhgCS4RRKbStq8AcW/4?Discord_ID=${user.id}&Username=${user.username}) to appeal your ban after you have [read our code of conduct](https://www.freecodecamp.org/news/code-of-conduct).`,
```

tested, it works only without that

majestic-owl448 (approved):
LGTM

PR #954, opened 2026-08-04
feat: allow staff to assign roles on behalf of users

### LINKED ISSUE #943: feat: allow staff to manage self-assignable roles on behalf of users
## Summary

Users are struggling to self-assign roles (particularly the AI role) using the existing self-serve flow. Staff currently have no way to assign these roles on a user's behalf through CamperChan.

## Requested Change

Add a staff-facing command or action in CamperChan that allows staff members to assign or remove self-assignable roles on behalf of a user who is unable to do so themselves.

✨ This issue was created with help from Hikari~ 🌸

majestic-owl448:
Checklist:

Closes #943

Adds `/management assign-role`, which toggles a role for a target member the same way the self-serve role buttons do. Moderators can now sort out campers who cannot work the onboarding UI.

The `role` option is autocompleted, and the list is computed live rather than configured: every role below the CamperChan's own highest role, minus the everyone role and minus roles owned by an integration, such as the booster role. Nothing to maintain when roles are added, moved, or removed. The same check runs again when the command executes, since a client can submit any string to an autocompleted option.

Autocomplete had no plumbing in this repository, so the first commit adds it: an optional `autocomplete` handler on the command and subcommand interfaces, plus a branch in `handleInteractionCreate` that dispatches to it. Every path there responds, because Discord accepts nothing but a choice list for these interactions and the camper's client hangs until timeout otherwise. The `management` command routes autocomplete through the same subcommand map as `run`, and runs the permission validator first, so non moderators get an empty list.

Role changes are posted to the mod webhook, in the same plain content style as the voice reports:

```md
<@moderator> has added the `AI` role to <@camper>.
```

Discord's audit log credits the CamperChan for the change, so without this there is no record of which moderator made it. The role is named rather than mentioned, so the log cannot ping everyone holding a mentionable role.

No configuration or permission changes are needed to deploy this. Commands are registered on boot, and the Manage Roles permission and the `GuildMembers` intent are both already in use.

### DIFF
--- a/src/commands/management.ts
+++ b/src/commands/management.ts
@@ -5,19 +5,39 @@ import {
   SlashCommandSubcommandBuilder,
 } from "discord.js";
 import { errorHandler } from "../utils/errorHandler.js";
+import { handleAssignRole } from "./subcommands/management/handleAssignRole.js";
 import { handlePrivate } from "./subcommands/management/handlePrivate.js";
 import { handleRole } from "./subcommands/management/handleRole.js";
 import { handleTickets } from "./subcommands/management/handleTickets.js";
 import type { Command } from "../interfaces/command.js";
 import type { Subcommand } from "../interfaces/subcommand.js";
 
+/* eslint-disable @typescript-eslint/naming-convention -- These keys are subcommand names, which Discord requires to be kebab-case. */
 const handlers: Record<string, Subcommand> = {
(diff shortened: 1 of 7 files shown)


naomi-freecodecamp (changes requested):
blocking while discussing on discord

